PR #31, opened 2017-12-11
Add GELF in/out plugin documentation

fbalicchia:


fbalicchia:
Hi, 
IMHO can be useful to user like me to know how debug Logagent in easy way. With this commit I'd like to share a way to do it.

otisg:
Probably better to keep separate stuff, like Logagent debug info, in a separate PR, but I merged it all together here.  Thanks @fbalicchia!

PR #35, opened 2018-01-24
added api-docs link to www docs. 

sivasamyk:
Added links to static api docs (will be active once Ciprian updates deployment)
Also added links to EU and US server swagger UI.

Once merged to master, will work with Ciprian to publish docs.

screen shot 2018-01-24 at 7 37 10 pm

### DIFF
--- a/docs/api/rest-api.md
+++ b/docs/api/rest-api.md
@@ -1,10 +1,39 @@
 
-Sematext Cloud exposes REST API to access some of its functionalities. 
+#### Introduction
+Sematext Cloud exposes REST API to access its functionalities. The REST API can be used to easily manage your Account and Apps remotely, without the need to be logged-in via a browser. It also makes it possible to automate certain tasks, such as creation of new Apps, definition of Alert Rules, etc.
 
-To know more about REST API go to <a href="/docs/rest-api" target="_blank">API Documentation</a>.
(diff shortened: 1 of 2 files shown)


otisg:
LGTM.  I suppose we could add redirect from the existing/old API pages under docs, but I think we need @hakman to make that happen in Apache configs?

hakman:
Deffinately worth a try. @sivasamyk can you create a Jira with the details?

sivasamyk:
@otis Do you mean the API Reference section under Log Management and Performance Monitoring sections in www.sematext.com/docs? Check the attached screenshot. The APIs listed there cover more functionality that that are exposed in V3. If we to redirect the API Reference under both Log Management and PM sections, we need to change the YAML under docs repo.
screen shot 2018-01-24 at 11 14 26 pm

otisg:
Yes, that is what I meant.  I didn't check what v3 covers and what those cover.  My understanding was that v3 includes old stuff, too.  If not, I think it should.  We don't want to have v2 in one place, v3 in another place, and each of them showing something different.
CC @bsmid

bsmid:
v3 consists just of selected endpoints, about 10 of them for now. v3 doesn't and shouldn't include old stuff and docs for v3 should be kept separate from v2 docs. We must make it clear that v2 docs are old and will be removed soon (add a note there). What Siva worked on should display only v3 API endpoints.

Separately, we should have someone go through v2 API docs and see which endpoints are mentioned there and don't exist in v3 yet and create v3 version for them. After that we can make v2 obsolete for good and forget we ever had it.

megastef:
Is this ready to merge/publish?

otisg:
Not just yet, @megastef but I think it will be next week.

sivasamyk:
@otisg @bsmid Update the API docs based based on content from www for existing v2 APIs. New page looks like:
screen shot 2018-02-16 at 2 30 09 pm

bsmid:
Looks good.

otisg:
LGTM @sivasamyk! 👍

PR #45, opened 2018-04-27
SC-1778: Update metric definitions in docs based on new metaInfo

sivasamyk:
Following are the columns:
`Metric Name                 |  Key                                                |  Metric Type  |  Numeric Type  |  Unit                                |  Description`

Generated for apps that have metainfo in genericConfig except HBase. 
There are some missing API metrics in spark.


otisg:
We've OK dropping the Agg column?

bsmid:
I think Aggregation function with our new logic becomes a property of report. Metric will have a type gauge or counter (for which report aggregation is implied to always be Avg and Sum, respectively). And report definition sometimes overrides this and defines different Agg function, but that doesn't affect the metric itself since user will be able to use any Agg function he wishes on it.

PR #46, opened 2018-05-11
GeoPoint: added short description of geo point

prog8:


otisg:
See my comment on the previous commit.  I think https://sematext.com/docs/logs/field-types would be better.  We can add just Geo Point there, and others later.

prog8:
It makes sense. I will do it on Monday then because it requires a little more content.

prog8:
@otisg @gr0 I added a short description of each field type

docs/logs/field-types.md
```diff
@@ -0,0 +1,106 @@
+You can see field types that you can chose in Field Editor when adding or modifying a field.
```
otisg: Note that we use Title: <title> as the very first line, followed by an empty line to tell Mkdocs to use that as the page heading and also in <title> tag.

docs/logs/field-types.md
```diff
@@ -0,0 +1,106 @@
...
+inside log shipper). You can use Analyzed String or Non-Analyzed String types,
+but note that this will prevent you from applying functions like min, max, avg, etc.
+on such field.
+A better option is to configure your log shipper to strip non-numeric portion and
+ship only numerical values. This will allow you to use this field for aggregations and reports.
+
+
+Example of a document containing a floating point number:
+
+    curl -XPOST https://logsene-receiver.sematext.com/<token>/example/ -d '{
```
otisg: Things like this we put in ``` blocks.

prog8: But this is code block and is interpreted as such (see the indent)

otisg: Oh, right, indenting does that, too, I forgot.  👍

docs/logs/field-types.md
prog8: I thought that inheriting the title from the file name ("Field types") is enough. Should I explicitly write "title: Field types". Is this what you mean?

otisg: Does inheriting make Mkdocs create a <title>Field types</title>?  If not, then explicit Title: Field types as the very first line is needed.

prog8: This is what it looks like. I can try to type "title ..." explicitly to see how it behaves

selection_376

prog8: I tried it and with "title: ..." it looks the same. Do you think that makes sense to duplicate the name and put it in 2 places (menu and title explicitly inside the file)? I am afraid that we may forget to change one place and I guess in 99% cases we need the same title in the menu and inside the file.

otisg: Key thing: just because you see a *heading* on a page doesn't mean that there is actually <title> tag in the page :) . So that's what you want to look for.

prog8: I thought you were referring to the heading but there is also `<title>` tag  in HTML if this is what you mean

otisg: For future reference.  There always is a <title> tag in the generated HTML.  The question is just what Mkdocs put in it.  Think Google SERPs, think SEO.  It used to put some default until we figured out we could use the Title: "trick" to tell it which exact <title>....</title> to use.

prog8: Mkdocs put there a title of the menu element. I really checked this.

prog8:
Waiting for green :traffic_light:  (approve).

PR #62, opened 2018-10-16
Added reserved tags section in tags doc

sivasamyk:


docs/monitoring/tag-support.md
```diff
@@ -3,18 +3,29 @@ description: Extract tags from metric sources, environment, assign metadata to h
...
+
+These tag names are used for internal purposes:
+
+| Tag Name  | Description  |
+|:--|:--|
+| os.host | Hostname of the host where the agent is running |
+| token | Sematext App Token |
+| container.hostname | Hostname of the container being monitored |
+| container.host.hostname | Hostname of the host where the container is running |
+| measurement | Used in Influx Line Protocol. Represents the top namespace to which the metrics from the Influx line belong |
```
bsmid: This might sounds a bit confusing to users, they may understand it as something they have to provide. Perhaps just say it is reserved name per Influx protocol and can't be used in tag names.

docs/monitoring/tag-support.md
```diff
@@ -3,18 +3,29 @@ description: Extract tags from metric sources, environment, assign metadata to h
...
+### Reserved tags
+
+Below tag names are reserved by Sematext to be used for internal purposes or for future use. It is not recommended to use these tag names for custom tags.
+
+These tag names are used for internal purposes:
+
+| Tag Name  | Description  |
+|:--|:--|
+| os.host | Hostname of the host where the agent is running |
+| token | Sematext App Token |
```
bsmid: If AA docs will point here, we need to tell that token can also be any custom string in right 36-char format when sending data elsewhere (not to SC)

docs/monitoring/tag-support.md
```diff
@@ -3,18 +3,29 @@ description: Extract tags from metric sources, environment, assign metadata to h
...
-create more useful dashboards, you can assign metadata to each
+To help you manage your metrics, hosts, and containers, and to help you create more useful dashboards, you can assign metadata to each
 host/server/container in the form of *tags*.
 
-Tags let you organize your hosts/servers/containers in different
-ways – for example by role, owner, or environment. Each tag consists of
-a key and a value, separated by the ':' character. Both key & value are
-case-sensitive.
+Tags let you organize your hosts/servers/containers in different ways – for example by role, owner, or environment. Each tag consists of
+a key and a value, separated by the ':' character. Both key & value are case-sensitive.
```
bsmid: We probably have to mention tags automatically added by agents based on monitored metric nature.

PR #65, opened 2018-11-27
SC-3625 www docs adjustments for token/rec rename

bsmid:
Rename of:
- app-token into monitoring-token
- setup-spm into setup-sematext

These are docs specific to "our" spm-client (not the public one) so changes should be good (they don't fully apply to public AA though).


megastef:
When can we  published the change?

bsmid:
I planned to publish in the next day or two. New spm-client is already deployed to prod.

PR #70, opened 2019-01-22
Fix to add anchor links to FAQ

sivasamyk:
- Explicitly add anchor tags with name and href to the FAQ. This way the FAQ questions is not present in ToC and we can have relative hyperlinks to individual FAQs
- Hide hyperlink styling for FAQ
- Observed on opening the relative link to an FAQ, the question is slightly hidden on top, but the content is visible completely.


otisg:
👍 You changed this manually or with some script?

sivasamyk:
Its a combination of both manual and script. Used script to extract and convert titles to id/names & manually replaced `**` in title with anchor tags (this could also be done using sed)

PR #97, opened 2019-06-07
SC-4132: Tag docs updates

sivasamyk:
- Renamed `Related special tags` to `Synonymous Tags`
- Minor changes to reserved special tags and move error related tags to Logs section

### DIFF
--- a/docs/monitoring/tag-support.md
+++ b/docs/monitoring/tag-support.md
@@ -100,7 +100,7 @@ Below are the tags that are reserved for future use.
 | Tag Name  | Description  | Synonymous Tags
 |:--|:--|:--
 | os.host.ip | IP Address of the host/server | server.ip, server.address, host.ip, host.address, source.ip, source.address
-| service.name | Name of the service where the metrics/logs is collected from |
+| service.name | Name of the service where the data is collected from |
 | service.id | Unique service identifier |
 | service.type | Service type e.g. `hadoop` |
 | span | Building block of trace in distributed tracing |
--- a/docs/monitoring/tag-support.md
+++ b/docs/monitoring/tag-support.md
@@ -14,9 +14,9 @@ The following tags are treated as special in Sematext Cloud and cannot be used i
 
 The tags below are applicable to all metrics types:
 
-| Tag Name  | Description  | Related special tags
+| Tag Name  | Description  | Synonymous Tags
 |:--|:--|:--
-| os.host | Hostname of the host where the agent is running | host, hostname, host.id, host.ip, host.name |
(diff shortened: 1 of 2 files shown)


docs/monitoring/tag-support.md
```diff
@@ -97,12 +97,13 @@ Below are Serverless related tags sent as part of metrics/logs in the Serverless
 
 Below are the tags that are reserved for future use.
 
-| Tag Name  | Description  | Related special tags
+| Tag Name  | Description  | Synonymous Tags
 |:--|:--|:--
-| source |  Represents the source of the event  | facility
-| service | Service that generated the event | service.id, service.name
-| span | Building block of trace in distributed tracing | 
-| error.code | Error code | error.id, error.message
```
bsmid: Do you plan to add this one somewhere else? Since it might be common for Logs and Metrics (and RUM? and Tracing? and...), maybe we need some doc for these common tags.

sivasamyk: I added the error fields in the logs section.  All tags listed here might be common across Sematext products since they will be used for correlation across products. Shall we have top-level doc called `Sematext Special Tags/Fields` and add all the special tags from Logs, Monitoring (and RUM, Tracing ) there?

bsmid: Hm, maybe we should instead have docs/tags and describe all tags there?

> All tags listed here might be common across Sematext products

Yeah, but will users realize that if we have different doc pages each under different tree (like in this case, under monitoring)? Not sure how to best present this to users.

sivasamyk: > Hm, maybe we should instead have docs/tags and describe all tags there?
Yes. Top level doc page for special tags and fields. This page will only list the tags and fields. Instructions to edit tags & tag aliases will remain in respective product pages.

We could list all special tags & fields here and provide a link to this page from monitoring, logs, RUM, etc. This way user will have a single place to look at all special tags/fields.

otisg: FYI @adnanrahic .  I see mentions of docs "refactoring".

adnanrahic: I like the idea. But, let's wait for the next iteration of Docathon to add this. I want to wrap up these docs and add Process and Inventory first, then publish it ASAP. After that, I'm eager to jump on new stuff like a dedicated Tags page.

sivasamyk: OK. Then I will merge this PR. I have started working on separate tags doc. I will merge this in the next iteration.

adnanrahic: Awesome! Thanks Siva. :smile:

PR #107, opened 2019-07-02
IN-1879: added more guides to the infrastructure monitoring docs

adnanrahic:
I've expanded on @bsmid 's initial docs for Infra with two screen captures and more explanations.

Screenshot below.
infra-monitoring-docs

### DIFF
--- a/docs/monitoring/infrastructure.md
+++ b/docs/monitoring/infrastructure.md
@@ -55,7 +55,7 @@ Creating a new Infra App is the same process as with any App in Sematext. In the
   <source src="https://sematext.com/wp-content/uploads/2019/07/create-infra-app-1.mp4" type="video/mp4" />
 </video>
 
-You can then choose which operating system your machine are running. There are two simple setup commands you need to run. After that you're ready to go.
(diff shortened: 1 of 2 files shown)


docs/monitoring/infrastructure.md
```diff
@@ -37,15 +38,38 @@ When doing a clean installation, regardless of whether you're installing in cont
...
+
+## Create an Infra App
+
+Creating a new Infra App is the same process as with any App in Sematext. In the Integrations view you choose Infra App, give it a name and click continue.
+
+<video style="display:block; width:100%; height:auto;" controls autoplay loop>
+  <source src="https://sematext.com/wp-content/uploads/2019/07/create-infra-app-1.mp4" type="video/mp4" />
+</video>
+
+You can then choose which operating system your machine are running. There are two simple setup commands you need to run. After that you're ready to go.
```
bsmid: Maybe instead "You can then choose..." be more precise about what is user actually doing here (otherwise it may sound like user is choosing some settings)? For example: "Next screen shows installation instructions which you'll have to follow to start using Infra App. Select the operating system running on your machines and run simple setup commands..."

adnanrahic: Edited.

PR #109, opened 2019-07-03
Wording

otisg:
Improved wording, but still don't like it.  I'll leave a comment in a bit.

### DIFF
--- a/docs/logagent/output-filter-removefields.md
+++ b/docs/logagent/output-filter-removefields.md
@@ -4,14 +4,14 @@ Description: GDPR, log anonymizer, reduce log volume
 
 ## Output filter: remove-fields
 
-This plugin removes fields before the output happens. All occurrences of the original field values can be replaced in the log "message" field with the string `!REMOVED!`. To mask the field values in the message field set the flag `maskValuesInMessageField` to `true`. 
+This plugin removes fields before sending them to the output destination. Alternatively, instead of completely removing fields and their values completely, this plugin can replace all occurrences of the original field values with string `!REMOVED!`. To mask field values instead of removing them set the `maskValuesInMessageField` flag to `true`. 
 
-In the context of data protection regulations like GDPR you might need to mask data fields, especially when you handover log data to 3rd parties. 
+In the context of data protection regulations like GDPR you might need to mask data fields, especially when you hand over log data to third parties. 
 
 
 ### Configuration 
 
-Add the following section 'outputFilter' to the Logagent configuration file. Please note you could use the plugin with multiple configurations for different event sources. 
+Add the following 'outputFilter' section to the Logagent configuration file. Note that you can use the plugin with multiple configurations for different event sources. 
 
 ```yaml
 # tail web server logs
@@ -39,7 +39,7 @@ Run Logagent with your config:
 logagent --config logagent-example-config.yml -n httpd --yaml
 ```
 
-The output does not contain client_ip and user field: 
+The output does not contain client_ip and user fields: 
 
 ```
 logSource:    httpd


docs/logagent/output-filter-removefields.md
```diff
@@ -4,14 +4,14 @@ Description: GDPR, log anonymizer, reduce log volume
 
 ## Output filter: remove-fields
 
-This plugin removes fields before the output happens. All occurrences of the original field values can be replaced in the log "message" field with the string `!REMOVED!`. To mask the field values in the message field set the flag `maskValuesInMessageField` to `true`. 
+This plugin removes fields before sending them to the output destination. Alternatively, instead of completely removing fields and their values completely, this plugin can replace all occurrences of the original field values with string `!REMOVED!`. To mask field values instead of removing them set the `maskValuesInMessageField` flag to `true`.
```
otisg: @megastef 

> To mask field values instead of removing them set the `maskValuesInMessageField` flag to `true`. 

I don't like this part. This `maskValuesInMessageField` makes it sound like this is about masking values on the *message* field.  But IIUC that is not correct.  This flag will just tell Logagent to replace values of specified fields with `!REMOVE!` instead of outright removing them.  In other words, this has nothing to do with "message" field.  Is this correct?  If yes, then the name of this property is misleading.

Separately from that, I don't like that we assume that `!REMOVED!` is OK by everyone.  Why not have a property that, if specified, means "mask field values and use <user-specified string> as the mask"?

e.g. something like:
maskFields: password, creditCard    # fields to mask
maskValue: XXXX                               # value to use for masking

megastef: > This maskValuesInMessageField makes it sound like this is about masking values on the message field. 

This is exactly what it does!

1) Removing the fields completly
2) Replace the field values in the message field (now optional via maskValuesInMessageField=true)
 
GDPR case:
Imagine you want to remove the email field, but the email address also was part of the message string, then you must 
a) Remove the email field (which was potentially extracted from original logline) 
   The field will be completely removed from the document (so it needs no disk space)
b) Mask each occurrence of the email address in the message string to avoid the transmission of personal data to a 3rd party. 
The mask `!REMOVED!` shows that there was a modification/processing. Of course, we could make the '!REMOVED!' mask configurable ... 

Please note we have other plugins to mask fields, those replace e.g. original value with hash codes or encrypted strings. See https://sematext.com/docs/logagent/output-filter-aesencryptfields/

otisg: But remove *field* is not the same as remove *value matching an expression*.
a) above is about removing *fields* (not from message field, but separate fields and their values)
b) is about stripping or masking strings that match the specified expression

Maybe this plugin does both a) and b)?
If so, then we can further adjust the doc to make that clear.

megastef: It does a) and b) for the GDPR use case. 

I would suggest: 

## Output filter: remove-fields

This plugin removes fields before sending them to the output destination. 

In the context of data protection regulations like GDPR, it might be required to avoid the transfer of personal data to 3rd parties. The removal of the data fields is one way to remove personal data in structured logs.

Often field values appear not only in one place. We might also find the personal data in the message field. Therefore, this plugin can mask the removed field values with the string `!REMOVED!` in the message field. To mask the field values in the message field set the `maskValuesInMessageField` flag to `true`. 

Example: 

The log parser generated following log event based on the log message `E-Mail delivery to user@example.com failed`.  

```
{"email": "user@example.com", status: "failed",  message: "E-Mail delivery to user@example.com failed"}
```

The result after removal of the email field and masking the actual email adress (field value): 

```
{"status": "failed", "message": "E-Mail delivery to !REMOVED! failed"}
```

The example shows that the plugin did two things:
1) The field "email" was removed
2) The field value `user@example.com` was replaced with `!REMOVED!` in the message field

otisg: OK.  One more question then.  Why is the message field assumed/hard-coded?  Why not have `maskValuesInFields`=field1,field2...

And give more control to the user.

megastef:
Closing this PR in favor of https://github.com/sematext/docs/commit/74f60520688e5b8934dadee4182fefda9542abfc

PR #117, opened 2019-07-15
Adjust url groups doc with new form

amir-hadzic:
Adjusted url groups page with information about the new Time Threshold field. Screenshot of changes below.

localhost_8000_experience_url-groups_


otisg:
@amir-hadzic can't we use "URL" everywhere consistently? (not url, Url, and such).

amir-hadzic:
Update the labels in UI and adjusted the screenshot.

PR #118, opened 2019-07-24
Fixed a documentation of geo fields

prog8:
As I wrote in JIRA it is not possible to detect geo fields in ES. The only geofield we support in our default mappings is called geoip.

### DIFF
--- a/docs/logs/field-types.md
+++ b/docs/logs/field-types.md
@@ -111,4 +111,4 @@ Example of a `geo` field which has geo-point type:
 
 If you want to use this field type you can send events with a field name and
 data in a format presented above. Alternatively, you can use Field
-Editor to manually create a field of geo-point type.
+Editor to manually create a field of the geo-point type.
--- a/docs/logs/field-types.md
+++ b/docs/logs/field-types.md
@@ -109,6 +109,6 @@ Example of a `geo` field which has geo-point type:
 
 **Note:**
 
-If you want to use this field type you can send documents with a field name and
+If you want to use this field type you can send events with a field name and
 data in a format presented above. Alternatively, you can use Field
 Editor to manually create a field of geo-point type.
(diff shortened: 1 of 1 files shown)


docs/logs/field-types.md
```diff
@@ -97,17 +97,18 @@ can be useful on dashboards presenting maps, for example.
...
       "message": "Hello from Sematext!"
     }'
 
 
 **Note:**
 
-If you want to use this field type you can send documents with a field
-in a format presented above. Data in such a format will be
-automatically detected as geo-point. Alternatively, you can use Field
+If you want to use this field type you can send documents with a field name and
```
otisg: documents => events

docs/logs/field-types.md
```diff
@@ -97,17 +97,18 @@ can be useful on dashboards presenting maps, for example.
...
     }'
 
 
 **Note:**
 
-If you want to use this field type you can send documents with a field
-in a format presented above. Data in such a format will be
-automatically detected as geo-point. Alternatively, you can use Field
+If you want to use this field type you can send events with a field name and
+data in a format presented above. Alternatively, you can use Field
```
adnanrahic: you can use Field => you can use *__the__* Field

docs/logs/field-types.md
```diff
@@ -97,17 +97,18 @@ can be useful on dashboards presenting maps, for example.
...
 Editor to manually create a field of geo-point type.
```
adnanrahic: Editor to manually create a field of geo-point type.
=>
Editor to manually create a field of _**the**_ geo-point type.

docs/logs/field-types.md
prog8: Are you sure that we should put the before "Field Editor"?

adnanrahic: Yeah.

prog8: We don't use "the" before Field Editor anywhere else, @otisg should be adjust all the places to have "the"?

otisg: Either way

PR #124, opened 2019-08-15
IN-5129: Added Experience integrations to main Integrations docs page

adnanrahic:
Adding the same Experience integrations to the main Integrations page in the docs.

### DIFF
--- a/docs/integration/index.md
+++ b/docs/integration/index.md
@@ -574,7 +574,7 @@ add geoIP information on the source IP addresses.
 	</div>
 </div>
 
-All of the integrations require adding and [configuring the Experience script](../experience/getting-started). However, single-page applications require one more configuration step to register route changes.
+All of the integrations require adding and [configuring the Experience script](../experience/getting-started). However, single-page applications require [one more configuration step](../experience/integrations) to register route changes.
 
 ## Notification Hooks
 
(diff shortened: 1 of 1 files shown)


docs/integration/index.md
```diff
@@ -473,6 +473,108 @@ target="_blank" rel="noopener">another AWS Lambda function</a>. If
...
+							<h5>Server-side rendered websites</h5>Full support for all types of SSR websites and applications.
+						</div>
+					</div>
+				</div>
+			</div>
+		</a>
+	</div>
+</div>
+
+All of the integrations require adding and [configuring the Experience script](../experience/getting-started). However, single-page applications require one more configuration step to register route changes.
```
otisg: Should we add an integration for "static websites"?

adnanrahic: I'd keep it here as part of the Experience integrations. What's your take on it? We can remove the SSR and Static Websites if you think they're not relevant.

otisg: I'm not sure what you are commenting on here.  I was asking if you think we should mention that this integrates with static websites, too, not just React, Vue.js, and friends.

adnanrahic: Oh! I didn't understand what you meant then. Yes, above this paragraph I've added the cards and Static Websites is among them. Check it out. :point_down: 
Selection_534

PR #127, opened 2019-08-28
IN-5311: add experience to top level docs and make cards prettier

adnanrahic:
I've added copy about Experience, and a new card for it. The cards are a bit prettier now as well. I've put them at the top for better visibility.

docs-top-level


adnanrahic:
Fixed! Merging and deploying now.

PR #138, opened 2019-11-09
Turned various container events to headings

otisg:
For better visibility in TOC


otisg:
Phew, thanks for checking (I don't have the env for docs set up locally)

PR #141, opened 2019-11-15
SC-5098 - Update STE docs

hakman:
STE docs needed an update for some time. This is a simplified version the k8s version in mind.


docs/sematext-enterprise/index.md
```diff
@@ -3,52 +3,36 @@ description: Sematext Enterprise is on-premises version of our managed SaaS plat
...
-**IMPORTANT:** The DEB and RPM packages have dependencies that will
-  automatically be fetched and installed.  If your system is behind a
+Given enough resources, such as RAM, disk, and CPU, it is possible to
+run all these services on a single host, although this is not recommended for
+a production setup. The number of worker nodes will depend on the number
+of monitored servers and size of collected logs. For high availability
+a minimum of 3 worker nodes is required.
+
+**IMPORTANT:** The Helm chart has Docker images as dependencies that will
+  automatically be pulled and run. If your system is behind a
```
otisg: will automatically be pulled and run => are are pulled and run automatically

mkdocs.yml
```diff
@@ -257,7 +257,6 @@ pages:
       - Overview: sematext-enterprise/index.md
       - Licensing: sematext-enterprise/licensing.md
       - On-premises vs SaaS: sematext-enterprise/on-premises-vs-saas.md
-      - FAQ: sematext-enterprise/faq.md
```
otisg: Why delete the FAQ?  Some of the questions on https://sematext.com/docs/sematext-enterprise/faq/ are still valid despite all the k8s stuff.

hakman: The FAQ is now part of the guide and is more technical. I don't think it makes sense to have them in 2 places
There are just a few Q that still seem relevant:
* Can I install Sematext Enterprise if my server is behind a firewall
* What server is used for sending emails

otisg: +1 for keeping the first one as that is something some people may wonder while exploring.  -1 for keeping the second one because that's something one thinks about only when actually setting things up.
But the point is we shouldn't remove the FAQ page all together.  For example, a good entry to add would be how STE and SC are different.  Or how does one upgrade STE.  Or what is EOL policy, and such.

docs/sematext-enterprise/index.md
```diff
@@ -3,52 +3,36 @@ description: Sematext Enterprise is on-premises version of our managed SaaS plat
...
-14.04), DEB package (Ubuntu 14.04 and 16.04), or RPM package (RHEL /
-CentOS 6 and 7) and contains the basic setup for all services that
-make up Sematext Enterprise.
+Sematext Enterprise is available as a Helm Chart (package for Kubernetes) and
+can be installed as such.  This will install everything, including components
+that have persistent storage.
 
 Given enough resources, such as RAM, disk, and CPU, it is possible to
 run all these services on a single system although that is not
 recommended for a production setup.
```
adnanrahic: This can be deleted because it is explained in the section below.

mkdocs.yml
adnanrahic: I agree. Keep the FAQ.

hakman: @otisg there is already a separate page for STE vs SC: https://github.com/sematext/docs/blob/master/docs/sematext-enterprise/on-premises-vs-saas.md

Regarding upgrades, it is an unknown to me also.
Do we have an EOL policy?

adnanrahic:
@otisg @hakman What's the state of this, when can/should we have this merged?

hakman:
@adnanrahic Done

PR #142, opened 2019-11-28
Fix issue reported by the user

gr0:
@megastef user on Hubspot reported that we use `IGNORE_LOGS_PATTERNS` instead of `IGNORE_LOGS_PATTERN`. This PR fixes it.


megastef:
Logagent uses `IGNORE_LOGS_PATTERN`:  https://github.com/sematext/logagent-js/blob/9b72ed05a4f0ea9d6245e7134d1feffdfa1a9ed0/lib/plugins/input/docker/docker.js#L21

PR #151, opened 2020-01-06
Make FAQ sections linkable

adnanrahic:
With this addition, every title in the various FAQ sections will be linkable.


adnanrahic:
They should continue to work fine. Mkdocs generates links for every heading in the FAQs where the heading text is added as a `#heading-text` to the URL of the page it's on. As long as the URL is properly formatted, everything should work fine. I've done checks of any links to any of these FAQ pages. And I fixed what I found. We should be good. Also, because we didn't have linkable FAQ headings, we didn't link them from anywhere in SC AFAIK, but I'll check this with Ciprian once again.

PR #155, opened 2020-01-13
Clarified that both name & identifier are required

otisg:


otisg:
@amir-hadzic merging this without approval

PR #159, opened 2020-01-21
SC-6750: merge tags and tag aliases

adnanrahic:
I've created a dedicated top-level section for tags that contains:

- Overview (general info about tags)
- Custom Tags (previously a part of the tag aliases page) 
- Special Tags

We have no more mentions of Tag Aliases in the docs.

screencapture-localhost-8000-tags-special-tags-2020-01-21-14_57_35
screencapture-localhost-8000-tags-custom-tags-2020-01-21-14_57_25
screencapture-localhost-8000-tags-2020-01-21-14_57_13


docs/tags/custom-tags.md
```diff
@@ -0,0 +1,50 @@
+title: Custom Tags 
+description: A Tag is an attribute of a data point (or metric) that could be used for grouping and filtering. The Sematext Agents support configuring custom tags. They can be specified in the Agent's configuration files. For example, tags like `env:prod` can be configured on all production servers.
+
+The Sematext Agents support configuring custom tags. They can be specified in the Agent's configuration files. For example, tags like `env:prod` can be configured on all production servers and `env:dev` on all development servers. The data can then be filtered in the UI using these tags. The maximum allowed length for both name and value is 1024 characters. They are mapped to `os.host` tag by the Agent. Configured tags are optional and can be changed at any time. The name and value of custom tags should match this regex: `[a-zA-Z0-9_\-=\+\.]*`.
```
otisg: I wonder if we should have Tags > Specs page where we show the list of tag specs/limits.  Like max length, allowed characters, max number of tax per ____, something about case sensitivity, etc.

mkdocs.yml
```diff
@@ -258,12 +258,15 @@ pages:
         - Configuration: sematext-docker-agent/configuration.md
         - Troubleshooting: sematext-docker-agent/troubleshooting.md
         - FAQ: sematext-docker-agent/faq.md
+    - Tags:
+      - Overview: tags/index.md
+      - Custom Tags: tags/custom-tags.md
+      - Special Tags: tags/special-tags.md
```
otisg: common-schema.md?

docs/tags/index.md
```diff
@@ -0,0 +1,56 @@
...
+
+At the moment all Common [Special Tags](../special-tags) apply to Monitoring Apps.
+
+If you choose to use 3rd party Agents, then you need to ensure the tag names match those from our Common Special Tags.
+
+## Automatic Tags collected from Agents
+
+The Sematext Agent automatically collects the following tags and sends them periodically to Sematext Cloud. It is not recommended to use these names for your own, custom configured tags.
+
+### Cloud Metadata
```
otisg: Wondering if, organizationally speaking, a separate cloud-tags.md would be easier for people?

docs/tags/index.md
```diff
@@ -0,0 +1,56 @@
...
+3. GCE - In GCE user-defined tags are called labels. To read labels, the instance needs `roles/compute.viewer` IAM role.
+    See [Granting Roles to Service Accounts](https://cloud.google.com/iam/docs/granting-roles-to-service-accounts#granting_access_to_a_service_account_for_a_resource) for more info.
+
+Cloud tags collection is enabled by default.  To disable Cloud tags
+collection set `cloud.metadata-enabled` to `false` in `/opt/spm/properties/st-agent.yml` and
+restart spm-monitor using `sudo service spm-monitor restart`.
+
+### Machine Metadata
+
+Following tags are collected from the host the agent is running on. They're mapped to the `os.host` tag.
```
otisg: Following tags are collected from the host the agent is running on.
=> 
Following tags are collected from the host the Sematext Agent is running on.
(and link the STA, so one knows the tags below apply only in that case?)

docs/tags/index.md
```diff
@@ -0,0 +1,56 @@
+title: Tags 
+description: A Tag is an attribute of a data point (or metric) that could be used for grouping and filtering.
+
+Tags are sent by Sematext Agents as part of every data point and they are shown in UI as filters. 
+
+For example, as part of Docker container metrics, the agent sends hostname on which the container is running, container identifier, container name, container image as tags which appear as filters in UI. User can then group or filter the container metrics using these tags.
+
+At the moment all Common [Special Tags](../special-tags) apply to Monitoring Apps.
```
otisg: We'll have to see what this feels like when published.  I see we have "Special Fields" under /docs/logs, and here we have Special Tags under /docs/tags.  A bit of asymmetry there...

docs/tags/custom-tags.md
```diff
@@ -0,0 +1,50 @@
...
+
+Below are the steps to configure custom tags in Sematext Agents.
```
bsmid: Using plural in "Sematext Agents" may be confusing. Sematext Agent generally refers to STA.

docs/tags/custom-tags.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Sematext Agent
+
+To add tags in the Sematext Agent, edit the property below in the Infra Agent configuration file:
```
bsmid: "add tags in" -> "add tags to"? Also, what is now "Infra Agent"?

docs/tags/custom-tags.md
```diff
@@ -0,0 +1,50 @@
...
+`/opt/spm/spm-monitor/conf/spm-monitor-config-${token}-${jvm}.properties`:
+
+``` properties
+# add tags if you want to use them, example: SPM_MONITOR_TAGS="env:foo, role:bar"
+SPM_MONITOR_TAGS="env:dev, project:projectName, role:webfrontend"
+```
+
+### Defining Tags in Integration YAMLs
+
+In the App Agent, tags are extracted automatically from metric data sources & values based on YAML configuration. Tags can be configured in the `tag` section of the App Agent integration YAMLs. The maximum allowed length for the tag name is 200 characters. The tag name should match this regex: <nobr>`[a-zA-Z0-9_\-.:(\\ |,=)]+`</nobr>. Examples of these tags are hostname, port, webapp name, jvm name, disk, elasticsearch index, etc. You don't need to adjust these tags for built-in
```
bsmid: App Agent -> Sematext App Agent? That might be more consistent with the rest, otherwise one can ask what is the diff between "Sematext App Agent" and "App Agent".

docs/tags/index.md
```diff
@@ -0,0 +1,56 @@
...
+
+If you choose to use 3rd party Agents, then you need to ensure the tag names match those from our Common Special Tags.
```
bsmid: Agents probably shouldn't be capitalised.

docs/tags/index.md
adnanrahic: Hm yes agreed. @bsmid mentioned we should also have a tags section under Monitoring, which will link to the top-level tags page. I'll rename the Special Tags to Common Schema.

docs/tags/custom-tags.md
adnanrahic: I'm all for it, but don't know what to write in such a page. Who would know what exactly we should add to the Tags > Specs page?

docs/tags/index.md
adnanrahic: Hm, maybe. I'd put both Cloud and Machine tags on a page. Let me try something...

docs/tags/index.md
adnanrahic: Added

docs/tags/custom-tags.md
adnanrahic: Fixed

docs/tags/custom-tags.md
adnanrahic: Fixed. Yeah, "Infra Agent" is the old naming, I changed it to Sematext Agent everywhere but forgot to do so here.

adnanrahic:
screencapture-localhost-8000-monitoring-tags-2020-01-22-12_08_35
screencapture-localhost-8000-tags-common-schema-2020-01-22-12_08_10
screencapture-localhost-8000-tags-custom-tags-2020-01-22-12_07_57
screencapture-localhost-8000-tags-2020-01-22-12_07_46

adnanrahic:
I've split up the Tags > Overview page and put the top of the old Overview page on the Monitoring > Tags page, with a link to the Tags > Overview.

The rest has been edited per your review.

mkdocs.yml
adnanrahic: Fixed

docs/tags/custom-tags.md
adnanrahic: I've created a task to not lose track of it: https://sematext.atlassian.net/browse/SC-6875

docs/tags/index.md
adnanrahic: I've kept both Cloud and Machine metadata on the Tags > Overview, but shortened the page so that the focus is on those tags.

docs/tags/custom-tags.md
adnanrahic: fixed

docs/tags/index.md
adnanrahic: fixed

docs/tags/custom-tags.md
otisg: > Who would know what exactly we should add to the Tags > Specs page?

You should.  Bojan would.  I might.
Let's just do it now.  It's 3.5 lines of text.  Stick there what you know (it's stuff like what you can see above - max length, allowed chars), and we can expand from there if there is more to add.

adnanrahic:
> Monitoring > Tags -- mentions that Tags > Common Schema is for Monitoring Apps only.

Will fix to mention both Monitoring and Logs Apps.

> Why not add a "Logs Tags" or "Logs Fields" section to Common Schema, and just like under Monitoring > Tags we say a few words about tags and then link to Common Schema, we can do that from Logs > Special Fields (which we could rename to "Fields" later if it looks like that makes more sense given that "special" is moving to "common").

Good point. I'll make this happen.

> Monitoring > Tags -- mentions only STE, as if tags cannot be sent from AAs.

I feel as is I do not want to confuse customers with the fact we have multiple agents. It's all one package when you install it. The fact we have several things inside of the package is confusing and irrelevant. But, you are right. I'll make sure the wording is in plural.

> I know we are not exposing "Tag Aliases" anywhere, but should we be pointing out any differences between what we internally know are Tags and what we know are Tag Aliases? Should we be pointing out when one should define a tag one way and when another way?

I would need to check with @bsmid here, but AFAIK everything regarding Tag Aliases is now under Tags > Custom Tags.

docs/monitoring/tags.md
```diff
@@ -0,0 +1,12 @@
+title: Tags 
+description: A Tag is an attribute of a data point (or metric) that could be used for grouping and filtering.
+
+[Tags](../../tags) are sent by Sematext's Agents as part of every data point. Tags are shown in the UI as filters and group by elements. 
+
+For example, as part of Docker container metrics, the agent sends hostname on which the container is running, container identifier, container name, container image as tags which appear as filters in UI. User can then group or filter the container metrics using these tags.
```
otisg: User can then group or filter the container metrics using these tags
=> 
User can then group or filter the container metrics using these tags as well as build custom charts.
(link build custom charts to the new CB page, if that makes sense)

docs/monitoring/tags.md
```diff
@@ -0,0 +1,12 @@
...
+
+All Tags from the [Common Schema](../../tags/common-schema) apply to both Monitoring and Logs Apps.
```
otisg: Well,  not really ALL tags.  We have them grouped there.  So maybe we can just drop "All".
Does Experience have any "reserved" tags?  Or SM?  Could you please find out and add them if appropriate?  Or ask DRIs to send a PR?

adnanrahic: Fixed.

I've asked the boys, I'll get an answer tomorrow.

docs/monitoring/tags.md
adnanrahic: Fixed

docs/monitoring/tags.md
adnanrahic: Amir said Experience does not have any reserved tags. Siva said he'll check and get back to me tomorrow.

adnanrahic: I've added tags for SM in the Common Schema.

PR #160, opened 2020-01-22
autodisco stale resources removal

rabbitstack:
This PR adds a new section that explains how to remove stale AA containers and the internal Docker network created by agent.


bsmid:
So, user has to do this manually? No way of some remove-hook or shutdown-hook for STA where it would clean things? I think people will expect STA to clean stuff behind itself.

rabbitstack:
Well, this is what SC-6552 was about. It could be done by STA, even though this has some implications. That's why I was dropping this idea from the beginning.

- If user temporary stops/deletes or upgrades STA container, all running AAs would die. Hence, there would be a gap in metrics. It would also take time for STA to bring up AAs again.
- Removing AAs on agent's shutdown could clash with autodisco logic.

The problem is that we don't know when user if definitely deciding to give up on STA. I just don't like the idea that STA has to remove all resources on shutdown, but if we still want to go this road we can implement this logic in the next sprint.

bsmid:
It isn't nice to leave the garbage behind. It isn't nice to have "unstable" AAs or big load during STA upgrade and interfering with autodisco. Can't say which of the bad side-effects is worse...

docs/agents/sematext-agent/autodisco/removing-stale-resources.md
```diff
@@ -0,0 +1,30 @@
+title: Removing stale resources
```
adnanrahic: can you add a description here as well?

otisg:
> * If user temporary stops/deletes or upgrades STA container, all running AAs would die. Hence, there would be a gap in metrics. It would also take time for STA to bring up AAs again.
> * Removing AAs on agent's shutdown could clash with autodisco logic.

This sounds super scary and not quite right.  We should think of STA as something that _enables_ one to start AAs, but not necessarily be 100% their manager.  For example, if somebody stops AAs manually somehow, STA will not restart it.  So I would not think of AAs that remain running even after STA is stopped as garbage.

docs/agents/sematext-agent/autodisco/removing-stale-resources.md
```diff
@@ -1,5 +1,7 @@
 title: Removing stale resources
 
+description:  Removing AA containers/networks after uninstalling the agent
```
adnanrahic: Can you replace AA with App Agent, for it to make more sense? Many people don't know what AA is. Also, please check the whole document here and change AA to be App Agent.

adnanrahic: description:  Removing AA containers/networks after uninstalling the agent
=>
description:  Removing App Agent containers and networks after uninstalling the Sematext Agent.

docs/agents/sematext-agent/autodisco/removing-stale-resources.md
```diff
@@ -1,5 +1,7 @@
...
+
 In Kubernetes deployments, Sematext Agent will make sure to remove any lingering AA containers once the DaemonSet supervising Sematext Agent pods is deleted. Nevertheless, in vanilla Docker or Swarm setups you have to do the cleanup manually after uninstalling the agent.
```
adnanrahic: Same comment for AA.

PR #172, opened 2020-02-07
Added span* and trace* to Logs

otisg:
Added to Logs tags:
- span.id
- trace.id

Added span.id, trace, trace.id to reserved tags, though that is kind of redundant if we have the above four reserved by Logs, right?

Removed  reserved tags:
- span
- trace

And I assume we will be adding tags reserved for tracing, though we may not have anything we want to reserve just yet?


rabbitstack:
The tags ending with the `id` suffix have a better semantic accuracy in terms of propagation entities. I think we would like to get rid of `span` and `trace` tags.

PR #177, opened 2020-02-15
Added info about hitting field count limits

otisg:


otisg:
@megastef is everything correct?  Am I missing anything?

PR #187, opened 2020-02-20
Updated Tags pages

adnanrahic:
Note: Feedback welcome for (1) :smile: 

### 1. Explained Tags and Tag Aliases without mentioning Tags Aliases
screencapture-localhost-8000-tags-2020-02-20-15_33_29

### 2. Moved Machine and Cloud tags to Common Schema + Merged Machine tags with Operating System tags
screencapture-localhost-8000-tags-common-schema-2020-02-20-15_34_05

### DIFF
--- a/docs/tags/common-schema.md
+++ b/docs/tags/common-schema.md
@@ -41,14 +41,49 @@ The tags below are applicable to all metrics/logs types:
 
 ## Operating System Tags
 
-Below are the OS related tags sent as part of OS metrics/logs
+Below are the OS related tags sent as part of OS metrics/logs. They are collected from the host the [Sematext Agent](../agents/sematext-agent) is running on. They're mapped to the `os.host` tag.
 
 | Tag Name  | Description  | Synonymous Tags
 |:--|:--|:--
-| os.disk | Human readable name of the block device | 
-| os.disk.mountpoint | Mount point for the disk in the file system |
-| os.disk.fs.type | Type of the file system associated with the device |
-| os.network | Name of the network interface |
+| System Disk | os.disk | Human readable name of the block device | 
+| System Disk Mountpoint| os.disk.mountpoint | Mount point for the disk in the file system |
+| Filesystem Type | os.disk.fs.type | Type of the file system associated with the device |
+| System Network Interface | os.network | Name of the network interface |
+| System UUID | os.uuid | Unique ID based on SMBIOS specification |
+| OS Distribution Name | os.distro.name | Distribution name of the OS. e.g. `ubuntu` |
+| OS Distribution Version | os.distro.version | Version of the OS. e.g. `16.04` |
+| Kernel Version | os.kernel | Version of the Kernel. e.g. `4.4.0-130-generic` |
+| JVM Version | jvm.version | Version of JVM, if available in `PATH` |
+| Virtualization | virtualization | Virtualization Type. Possible values are `BareMetal`, `VM`, `Container` |
+| Container Labels | - | All user-defined container labels. These tag are mapped to `os.host` and `container.id` |
+
+### Cloud Tags
+
+The cloud metadata from AWS, Azure and GCE instances are collected as tags. They're mapped to the `os.host` tag.
+
+| Name  | Tag Name  | Supported Cloud Providers  |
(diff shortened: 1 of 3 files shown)


otisg:
Cloud Tags - should that really be indented like that?

I do think we can make the Overview better by educating a bit more about why tags are valuable instead of saying just that they are for simplifying SC usage.  I think we could also show where in SC they are shown and how to use them.  Maybe best to add screenshots/MP4 after the new top nav, though.

adnanrahic:
> Cloud Tags

Aha yes, now I see the indent is off.

I'll go in and add some more examples of tags with a screenrecording.

adnanrahic:
Added another paragraph and video. https://cdn.sematext.com/videos/groupbytags2.mp4

localhost_8000_tags_

otisg:
bq. Tags you define can be grouped and filtered...

Tags themselves are not grouped.  Data is grouped or filtered _using_ tags.

Is it really true that tags from the Common Schema cannot be updated?
What if I specify a tag with one of the tag names from the common schema?

adnanrahic:
> Is it really true that tags from the Common Schema cannot be updated?
> What if I specify a tag with one of the tag names from the common schema?

I'm not sure. @bsmid would know.

> Tags themselves are not grouped. Data is grouped or filtered using tags.

Aha yes, I'll re-phrase it.

bsmid:
> Is it really true that tags from the Common Schema cannot be updated?

I don't know what updating a tag would mean at all, there is no such feature.

I think the whole "Tags you define by yourself or tags defined on environment level..." should be removed. I am not sure what you tried to describe there?

docs/tags/common-schema.md
```diff
@@ -41,14 +41,49 @@ The tags below are applicable to all metrics/logs types:
 
 ## Operating System Tags
 
-Below are the OS related tags sent as part of OS metrics/logs
+Below are the OS related tags sent as part of OS metrics/logs. They are collected from the host the [Sematext Agent](../agents/sematext-agent) is running on. They're mapped to the `os.host` tag.
```
bsmid: "mapped to..." - not sure what you mean here?

docs/tags/common-schema.md
```diff
@@ -41,14 +41,49 @@ The tags below are applicable to all metrics/logs types:
...
+| OS Distribution Name | os.distro.name | Distribution name of the OS. e.g. `ubuntu` |
+| OS Distribution Version | os.distro.version | Version of the OS. e.g. `16.04` |
+| Kernel Version | os.kernel | Version of the Kernel. e.g. `4.4.0-130-generic` |
+| JVM Version | jvm.version | Version of JVM, if available in `PATH` |
+| Virtualization | virtualization | Virtualization Type. Possible values are `BareMetal`, `VM`, `Container` |
+| Container Labels | - | All user-defined container labels. These tag are mapped to `os.host` and `container.id` |
+
+## Cloud Tags
+
+The cloud metadata from AWS, Azure and GCE instances are collected as tags. They're mapped to the `os.host` tag.
```
bsmid: Again "mapped to" - one could imagine there is some sort of "mapping" in case of tag aliases, but since we don't want to differentiate between the tag types anymore, I think we shouldn't mention any relation to os.host.

docs/tags/index.md
```diff
@@ -1,46 +1,18 @@
 title: Tags 
 description: A Tag is an attribute of a data point (or metric) that could be used for grouping and filtering.
 
-The [Sematext Agent](../agents/sematext-agent) automatically collects the following tags and sends them periodically to Sematext Cloud. It is not recommended to use these names for your own, [custom tags](custom-tags).
+The [Sematext Agent](../agents/sematext-agent) automatically collects the following tags and sends them periodically to Sematext Cloud.
```
bsmid: Which following tags? We don't show tag types after this.

adnanrahic:
@bsmid I improved it a bit.

screencapture-localhost-8000-tags-2020-02-24-11_47_09

otisg:
Leaving to @bsmid to look at approve.  I'll approve now and if I see issues will correct directly, will be easier (for me).

PR #191, opened 2020-02-25
SC-7106: Added GitHub Webhook input plugin

adnanrahic:
Added plugin page for the GitHub Webhook input plugin.

### DIFF
--- a/docs/logagent/input-github-webhook.md
+++ b/docs/logagent/input-github-webhook.md
@@ -1,0 +1,88 @@
+title: Logagent input plugin for GitHub Webhook Events
+description: Logagent features modular logging architecture framework where each input or output module is implemented as a plugin. Logagent can receive GitHub Webhook events via HTTP. 
+
+## Input Plugin: GitHub Webhook Events
+
+Input plugin to receive GitHub Webhook Events via HTTP and send them to Sematext Events.
+
+Features:
+
+- List all crucial GitHub events in the Sematext Events stream
+
+Applications:
+
+- centralize GitHub Webhook Events
+- act as webhook to receive GitHub Webhook Events
+- index GitHub Webhook Events in Elasticsearch or Sematext Cloud
+- create alerts on GitHub Webhook Events in Sematext Cloud
+
+
+Requirements: 
+
+- configure the GitHub API to send Events via webhook 
+
+### Configuration
+
+```yaml
+# github.yml
+
+input:
+  github-webhook:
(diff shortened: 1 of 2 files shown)


docs/logagent/input-github-webhook.md
```diff
@@ -0,0 +1,88 @@
...
+Input plugin to receive GitHub Webhook Events via HTTP and send them to Sematext Events.
+
+Features:
+
+- List all crucial GitHub events in the Sematext Events stream
+
+Applications:
+
+- centralize GitHub Webhook Events
+- act as webhook to receive GitHub Webhook Events
```
otisg: Webhook vs. webhook

docs/logagent/input-github-webhook.md
```diff
@@ -0,0 +1,88 @@
...
+- index GitHub Webhook Events in Elasticsearch or Sematext Cloud
+- create alerts on GitHub Webhook Events in Sematext Cloud
+
+
+Requirements: 
+
+- configure the GitHub API to send Events via webhook
```
otisg: here, too

otisg:
Btw. can one send Github events to other destinations, too?

adnanrahic:
> Btw. can one send Github events to other destinations, too?

Not at the moment, no. I've tailor-made the input plugin to parse the data and structure it in a format that fits our Events. Basically, I've tailor-made this to be a receiver just like the journald receiver.

I'd need to restructure the input plugin to parse the data into a more general format and then have a dedicated output plugin that will parse that data into the format our Events need. This would mean that we need dedicated output plugins for every type of webhook, if we want every type of webhook input plugin to be able to send data to other destinations. Or, implement dedicated output filters that hook into a single output plugin. I'll check with @megastef .

megastef:
@adnanrahic I suggest doing the formatting work in an output-filter plugin. See .e.g https://github.com/sematext/logagent-js/blob/master/lib/plugins/output-filter/journald-format.js  - the input plugin does just little transformation by splitting bulk requests in multiple events, the output-filter is then formatting the data for a useful structure in Sematext Log Apps.

adnanrahic:
Aha, that's what I thought. Okay then, I'll think about how to go about this and create tasks accordingly. This can stay as-is for our receiver implementation.

I'll create new input and output plugins for general use cases.

adnanrahic:
@otisg I've created a task for myself to handle this: https://sematext.atlassian.net/browse/SC-7229

PR #197, opened 2020-03-20
SC-7068: added K8s audit logs integration page

adnanrahic:
I've added a page for the Kubernetes Audit Logs integration under integrations. I can move this page under Kubernetes when that page gets added.

localhost_8000_integration_kubernetes-audit-integration_ (2)

### DIFF
--- a/docs/integration/kubernetes-audit-integration.md
+++ b/docs/integration/kubernetes-audit-integration.md
@@ -38,6 +38,8 @@ metadata:
       burst: 15
     clientConfig:
       url: "https://logsene-k8s-audit-receiver.sematext.com/<LOGS_TOKEN>/"
+      # For EU Region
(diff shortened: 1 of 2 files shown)


otisg:
My only suggestion here is to add a sentence or two towards the end of the Quick Start paragraph that says there are 2 ways to configure things, so that it is clear that it is an either-or situation, and provide pros/cons/recommendation.  These 2 choices can then be linked to sections below - Dynamic Backend and the other one.

adnanrahic:
Good point. Let me add them as tabs, that might make more sense.

adnanrahic:
Sadly if I add tabs that will mess up the code syntax highlighting and make it look terrible. I left it as it was but added all the things you suggested.

screencapture-localhost-8000-integration-kubernetes-audit-integration-2020-03-20-17_39_48

PR #206, opened 2020-04-17
Removed an old, incomplete list of filters

otisg:
Removing this list because we have separate pages for input and output filters and we don't want to have this outdated and incomplete list here.

### DIFF
--- a/docs/logagent/filters.md
+++ b/docs/logagent/filters.md
@@ -132,10 +132,3 @@ outputFilter:
               }
             }
 ```
-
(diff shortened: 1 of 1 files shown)


otisg:
I assume this is correct, right @megastef ?

otisg:
I think they do work, but we don't need to list these 3 filters here when we have dedicated pages for lists of input and output filters.  In other words, I wasn't saying there is anything wrong with these filters, just that listing only 3 of them here is... meh.

PR #208, opened 2020-04-23
Added logpack docs and restructured integrations directory

adnanrahic:
I changed a lot and added LogPacks. Feedback is very welcome!

I cleaned up the Integrations Overview by putting integrations under respective folders. This does not impact the URL, it **ONLY CHANGES** the "folders" on the left.

I added a LogPacks section and added cards to be uniform with Monitoring and Experience. I also moved the content about logging to a dedicated page called `Generic Logs Integration`.

new-logs-ovreview

## Generic Logs Integration:

generic-logs-integration-docs

## Kubernetes Audit Integration:

k8s-audit-logpack-docs

## Github Webhook Events:

github-webhook-events-docs


otisg:
This looks great, Adnan!  We can tweak/update some wording here and there, but we can do that later once this is published.  The only hesitation here is that we are publicly mentioning LogPacks now and we should think if we really want to be introducing a new term like that or if we should just call this an Integration in public-facing docs...

adnanrahic:
I agree. Let's just go with Integrations. Everything is an _integration_ and I would like to keep the simplicity. Give me a moment to fix this.

adnanrahic:
Here's the updated copy. I think we can go with this for now. I'll merge and deploy this as soon as we merge the PR for this in SC.

PR #223, opened 2020-05-28
Update integrations with new instructions for TS and ESlint

amir-hadzic:
This fixes the integrations instructions to include recent updates for avoiding TypeScript and ESLint errors.

image


otisg:
Assuming this is correct, which I assume it is, this is good for merging and then publishing, @adnanrahic

PR #227, opened 2020-06-08
Add geoip section to Logs common schema

gr0:
Hi, 

This PR adds the Logagent `geoip` fields to common schema. During that I also changed the fields that the iOS and Android libraries are sending, so they are now sending the `geoip.location` just like Logagent does. This was done for consistency reasons. It was easier to change that in the libraries - less users, close to no issues with backwards compatibility.

### DIFF
--- a/docs/tags/common-schema.md
+++ b/docs/tags/common-schema.md
@@ -24,11 +24,10 @@ Refer to this list of default fields for Logs Apps. The tags below are applicabl
 | error.type | A reserved field for errors |
 | span.id | Building block of a trace in distributed tracing |
 | trace.id | Building block of a trace in distributed tracing |
-| geoip | An object reserved for location related data |
 | geoip.location | A latitude and longitude defining the location associated with the log line |
+| geoip.info.city | The city associated with the log line |
 | geoip.info.country | The country associated with the log line |
 | geoip.info.continent | The continent associated with the log line |
-| geoip.info.city | The city associated with the log line |
 
 All of these fields are optional, but their use is strongly encouraged. If found in logs with low-enough cardinality, all distinct values of these fields will be loaded and shown in the UI as filters and thus allowing one to very quickly narrow down the search.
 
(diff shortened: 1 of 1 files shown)


otisg:
> I also changed the fields that the iOS and Android libraries are sending

Doesn't that break backwards compatibility?  Should we add "Breaking Changes" to their READMEs?

gr0:
I checked - no one shipped data with geolocation from mobile devices, at least I don't see such data in our Logs indices. That's why I only added that information to release description:

- https://github.com/sematext/sematext-logsene-ios/releases/tag/1.4.0
- https://github.com/sematext/sematext-logsene-android/releases/tag/2.3.0

docs/tags/common-schema.md
```diff
@@ -24,6 +24,11 @@ Refer to this list of default fields for Logs Apps. The tags below are applicabl
 | error.type | A reserved field for errors |
 | span.id | Building block of a trace in distributed tracing |
 | trace.id | Building block of a trace in distributed tracing |
+| geoip | An object reserved for location related data |
+| geoip.location | A latitude and longitude defining the location associated with the log line |
+| geoip.info.country | The country associated with the log line |
+| geoip.info.continent | The continent associated with the log line |
+| geoip.info.city | The city associated with the log line |
```
otisg, line 31: May want to push this one up 2 rows.  city -> country -> continent.  Size/ordering thing.  OCD? ;)

docs/tags/common-schema.md
```diff
@@ -24,6 +24,11 @@ Refer to this list of default fields for Logs Apps. The tags below are applicabl
 | error.type | A reserved field for errors |
 | span.id | Building block of a trace in distributed tracing |
 | trace.id | Building block of a trace in distributed tracing |
+| geoip | An object reserved for location related data |
```
otisg, line 27: Do we need this at all?  Do we document objects or fields?  Doesn't seeing N fields with the same prefix imply "object"?

docs/tags/common-schema.md
gr0: Good point, adjusted.

docs/tags/common-schema.md
gr0: Looked around, we don't add objects. Removed.

gr0:
Thank you :)

PR #237, opened 2020-06-29
SC-8730 - Rename fields to use geo instead of geoip

gr0:
Hi @otisg - have a look at the rename :)


gr0:
I'll add those fields. I'm not sure if I have the data to fill those though, but I'll check.

PR #238, opened 2020-07-06
Docs john

johndemian:
I've implemented all the changes we've discussed and would like to go over them at your earliest convenience, Adnan.


adnanrahic:
@johndemian we should close this PR and open a new one after this is fixed.

PR #244, opened 2020-08-04
Added docs for 'field rename' Logagent output filter

adnanrahic:
Added a page under output filters:
localhost_8000_logagent_output-filter-renamefields_ (1)

Added an entry in the FAQ:
localhost_8000_logagent_faq_


adnanrahic:
Aha good point.

adnanrahic:
Done. :ok_hand:

PR #247, opened 2020-08-10
Linux logs integration docs

radu-gheorghe:


otisg:
Should this be a separate page or should the new bits be added to https://sematext.com/docs/integration/journald-integration/ ?

adnanrahic:
I think it should be a separate page as it has dedicated instructions in the UI. This way it's less confusing for customers.

If you think my logic is flawed, I'm open to feedback.

docs/integration/linux-integration.md
```diff
@@ -0,0 +1,63 @@
...
+<img
+  class="content-modal-image"
+  alt="Linux Logs Auth Report"
+  src="../../images/agents/linux_auth_logs.png"
+  title="Linux Logs Auth Report"
+/>
+
+Other built-in reports include:
+
+- **Kernel**: Logs filtered by the facility 0 (kernel). You will find here all your startup logs, information about crashes, all that you typically see via `dmesg`
```
adnanrahic, line 55: You will find here all your startup logs
=>
Here you will find all your startup logs

adnanrahic:
Aha so I have one idea that dawned on me now.

We can rename the Journald integration page to `Journald Integration HTTP`, and rename the Linux logs integration to Journald Integration HTTPS.

But then again this would need a rename in SC to have the naming consistent.

Hence, why I would like to have this named `Linux Logs Integration` and keep the old `Journald integration` page as is. 

CC @otisg @radu-gheorghe

otisg:
Leaving it to you, two.

radu-gheorghe:
@adnanrahic I've updated the integration page as we discussed: all 3 options in one, the recommended one on top (with `journalctl`). This depends on https://github.com/sematext/logagent-js/pull/252, so I guess we'll have to merge that one and release Logagent before merging this PR.

PR #255, opened 2020-08-25
SC-9318 - Add IP address to common schema

gr0:
This PR adds IP address to common schema.

### DIFF
--- a/docs/tags/common-schema.md
+++ b/docs/tags/common-schema.md
@@ -24,6 +24,7 @@ Refer to this list of default fields for Logs Apps. The tags below are applicabl
 | error.type | A reserved field for errors |
 | span.id | Building block of a trace in distributed tracing |
 | trace.id | Building block of a trace in distributed tracing |
+| geo.ip | IP address defining the location associated with the log line |
 | geo.location | A latitude and longitude defining the location associated with the log line |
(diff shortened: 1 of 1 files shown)


adnanrahic (approved):

docs/tags/common-schema.md
```diff
@@ -24,6 +24,7 @@ Refer to this list of default fields for Logs Apps. The tags below are applicabl
 | error.type | A reserved field for errors |
 | span.id | Building block of a trace in distributed tracing |
 | trace.id | Building block of a trace in distributed tracing |
+| geo.ip | IP address defining the location associated with the log line |
```
otisg, line 27: Does this make sense?  Does it make sense to add a field with a geo prefix?  That assumes the IP will be geodecoded.  Isn't there os.hostip?

gr0: I think it makes sense. It is not about the host IP. Think Apache logs. You are not interested in the IP of your host, but the IP of client that made the request.

adnanrahic: Yes, the hostip is not the public IP address. The public IP will either be taken from the logs (e.g apache) or will be looked up inside Logagent and then it will be the public IP of the host. It makes sense to me.

PR #260, opened 2020-09-01
Added info about changing the target Monitoring App

otisg:


docs/monitoring/spm-faq.md
```diff
@@ -378,6 +378,12 @@ bootstrap the JMX connector:
 
 See [enabling Security Manager permissions](/agents/sematext-agent/jmx-attaching/#security-manager-permissions) to allow automatic JMX connector exposure.
 
+### Can I reconfigure the agent to send metrics to a different Monitoring App?
```
bsmid, line 381: For that particular user this was the best approach. But there are possible variations:
- if in-process agent is used, then one also has to change token in their javaagent definition in startup script
- if autodisco is used, then just change the app in the UI

bsmid (approved):

PR #271, opened 2020-09-25
Added Mobile & Browser SDK

otisg:
I can't tell what colour these are, so if they are not good could you please change them?  Also, centering text may look better than having the text left aligned.


adnanrahic (approved):

rabbitstack:
Should we zap the "Docker Agent" card?

adnanrahic:
That card links to Sematext Agent > Docker Installation page. All good.

PR #276, opened 2020-10-07
SC-9435 - Add Signl4 notification hook to docs

gr0:
Hi guys, 

This PR adds Signl4 notification hook integration to docs and looks like this:

signl4


adnanrahic (approved):

adnanrahic:
This is good @gr0 , but please also add this notification hook to the docs/integration/overview.md in the `notification hooks` section.

gr0:
Good point, I again forgot about it. Will do.

gr0:
Actually, it is already there, I added it this time ;)

otisg:
In general - for this hook or others, I think we can skip the "create account" step.  If one is looking at this documentation then one already has an account with the other service.

docs/integration/alerts-signl4-integration.md
```diff
@@ -0,0 +1,30 @@
+title: Signl4 Alerts Integration
+description: Sematext threshold, anomaly and / or heartbeat Alerts integration with Signl4. Use them together to get the information about each notification quickly and on any device.
+
+## In Signl4
+
+**1.** Create a Signl4 account at [signl4.com](https://www.signl4.com).
```
otisg, line 6: I'd zap this here and in all other hook pages

adnanrahic:
Awesome! Good to merge then.

otisg (approved):

docs/integration/alerts-signl4-integration.md
gr0: Good point, I'll do that.

gr0:
Thanks guys!

PR #288, opened 2020-11-04
Split screen

emir-arnautovic:
Here is preview of split screen doc:
image

and changes in events:
image

and in monitoring

image


adnanrahic (approved):
I directly edited this PR. It's ready to be merged now.

otisg:
Nice PR!  Should anything be added to https://sematext.com/docs/logs/reports-and-components/ ?

Another thought: For each solution we could add "Correlation" to nav bar, have that point to a short page, say /docs/solution-here/correlation.md and on that page provide just a bit of info and a link to the main Split Screen page under /docs/guide/... Would that make sense?

emir-arnautovic:
We could add similar content in logs (and other "solutions") to point out that split screen can be used to correlate with other data.
I am a bit unsure about "correlate" pages because we will eventually add true correlations so we might want to save it for that.

otisg:
> I am a bit unsure about "correlate" pages because we will eventually add true correlations so we might want to save it for that.

When we do that when we can edit the content, add "true correlate" info, keep a blurb about split screen, etc.  i.e. this future stuff is not a blocker for us adding a correlate page under each solution today.  People looking under Logs docs will never discover Split Screen under Guide docs.

adnanrahic:
Good point. Emir, we could add a page called Correlation and just say a few sentences and point to the Split Screen page under User Guides. Do you want to do that?

otisg:
I was trying to point out that we need to put Correlation under *each* solution in the docs.  Because people who are looking for info about e.g. Logs will not look under Guides.

emir-arnautovic:
I get your point, but my point is that we used to have correlations so not sure if this should be called correlations. Or we add "Correlation" and we list now split screen and later we also add correlation feature?

otisg:
> we add "Correlation" and we list now split screen and later we also add correlation feature?

That. :)

emir-arnautovic:
Created SC-10059 for the next sprint.

adnanrahic:
Awesome. I'll merge this PR now as Split View is going to prod in SC today.

adnanrahic:
The Correlation docs will be added next sprint.

PR #294, opened 2020-11-17
Dont loop mp4s and don't autoplay unless video is in the viewport

mbonaci:
I intentionally did this in a separate branch.
Pls verify whether this is what we want.
I went through all the links locally to verify whether the video is in the viewport - I left autoplay only in that case.


adnanrahic (commented):
LGTM.

adnanrahic (approved):

PR #295, opened 2020-11-18
Add guide for saved views

amir-hadzic:
Here's the preview:

localhost_8000_guide_saved-views_


docs/guide/saved-views.md
```diff
@@ -0,0 +1,23 @@
+title: Saved Views Guide
+description: Sematext Cloud is a modern monitoring, log management, transaction tracing, and real user monitoring system that includes over 40 monitoring integrations. It is a suite of products that combine high-quality logging experience with other monitoring and alerting devops tools helping fix IT production issues
```
otisg, line 2: Should be about SVs, not generic.

docs/guide/saved-views.md
```diff
@@ -0,0 +1,23 @@
...
+
+Filtering and grouping data is a central part of the Sematext Cloud user interface. With Saved Views you can save your filters and groups for faster access later. If you find yourself often applying the same filters on your data then this feature can be a time saver.
+
+The feature is available on all screens where the Saved Views button is visible. This button displays "Default View" when no saved view is selected as shown in the animations below.
```
otisg, line 6: When I saw "Saved Views button" I wondered if we are showing this "Saved Views button" in the screenshot below.  But we don't.  Well, at least there is nothing that says "Saved Views", which is what I think a person would look for.  Since the button is not actually labelled "Saved Views", maybe we need to annotate the screenshot and point to it.

mkdocs.yml
```diff
@@ -12,6 +12,7 @@ pages:
       - Alerts Guide: guide/alerts-guide.md
       - Events Guide: guide/events-guide.md
       # - Dashboards Guide: guide/dashboards-guide.md
+      - Saved Views: guide/saved-views.md
```
otisg, line 15: Is Guides the right place?  Or would it make more sense to put it under Dashboards?

otisg (changes requested):

docs/guide/saved-views.md
amir-hadzic: I looked at few pages, and seems we use the same excerpt everywhere or maybe I hit the few pages that have the same. Will check...

docs/guide/saved-views.md
amir-hadzic: The screenshot is actually an animation that I think does a good job of showing the functionality, it's hard to miss it. Let me attach the animations here:

SavedViewSelect
SavedViewCreate

mkdocs.yml
amir-hadzic: This is available in different types of Apps and also on dashboards, similar to Split Screen which is in Guides. I could add a link to this from some part of Dashboards.

amir-hadzic: Added a link under Dashboards -> Reports and Components, similar to how I linked it from Monitoring.

docs/guide/saved-views.md
amir-hadzic: Updated the description to "Sematext Cloud is a modern monitoring, log management, transaction tracing, and real user monitoring system that includes over 40 monitoring integrations. Saved Views is a feature that allows saving filters and groups for faster access later."

emir-arnautovic: We've placed split screen under "User Guide":
image

I guess saved views and split screen are pretty much the same level.

mkdocs.yml
otisg: I think we are struggling with finding the best place for functionality that spans multiple solutions...  I don't have an immediate suggestion that feels great.  So I don't want to block this now.

docs/guide/saved-views.md
otisg: Ahaaaaa.  Problem solved then.

docs/guide/saved-views.md
```diff
@@ -0,0 +1,23 @@
...
+
+To save a view for later, follow the steps below:
+
+1. Change the filters according to which data you want to see
+2. Click on the highlighted Saved Views button
+3. Click on "Save As" and enter a descriptive name for the filter
+
+![Creating the saved view](../images/guide/savedviews/SavedViewCreate.gif)
+
+Sometimes later, when you come back to this screen:
```
adnanrahic, line 16: Sometimes later, when you come back to this screen
=>
Later, when you come back to this screen

docs/guide/saved-views.md
```diff
@@ -0,0 +1,23 @@
...
+
+1. Click on Saved Views button
```
adnanrahic, line 18: Click on Saved Views button
=>
Click the Saved Views button

mkdocs.yml
adnanrahic: Yes, we need to think of a better place to reference the Saved Views. For now, it is fine to have them under Guides. But we need to have them referenced for every solution.

adnanrahic (approved):
Two small edits and it's good to go.

PR #298, opened 2020-11-24
Rewrite on-page transactions as measurements

amir-hadzic:
This PR replaces On-Page Transactions docs page with Measurements.

The story behind this is that we are switching to using the native `performance.measure()` API instead of relying on our own terms and API for recording custom metrics in websites. We are hoping that this will be easier and less vendor lock-in than our previous approach.

image


mkdocs.yml
```diff
@@ -68,7 +68,7 @@ pages:
       - URL Groups: experience/url-groups.md
       - Configure Apdex: experience/configure-requirements.md
       - User Satisfaction: experience/user-satisfaction.md
-      - On-Page Transactions: experience/on-page-transaction.md
+      - Measurements: experience/measurements.md
```
otisg, line 71: Should this be called "Performance Measurements" for a bit more clarity?

otisg (changes requested):

docs/experience/measurements.md
```diff
@@ -0,0 +1,70 @@
...
+Click on a measurement in the table to see data related to that measurement only.
+
+
+## Custom Tags
+
+If you wish to apply custom tags to all measurements, then please check out [Tags](/experience/tags).
+
+If you want to add a specific custom tag to a single measurement, then you will have to use our custom commands for recording measurements instead of using the `performance.measure()` API.
+
+You can attach custom tags to measurements by providing them as another argument when calling `startTransaction`:
```
gr0, line 56: Maybe we can link to Tags page (https://sematext.com/docs/experience/tags/) for Experience here?

gr0 (approved):
Left one comment, apart from that looks good 👍

docs/experience/measurements.md
```diff
@@ -0,0 +1,70 @@
+title: Measurements
```
adnanrahic, line 1: Measurements => Performance Measurements

docs/experience/measurements.md
```diff
@@ -0,0 +1,70 @@
...
+performance.mark('custom-metric-end');
+
+performance.measure('custom-metric', 'custom-metric-start', 'custom-metric-end');
+```
+
+This will record a measurement with name `custom-metric` and the duration will be equal to the time it took for `doSomePotentiallyLongWork()` function to return.
+
+This measurement will be reported back to Experience every time this code runs, where you will be able to see the average duration for all users, compute the user satisfaction score, filter the data and more.
+
+### Setting The Time Threshold
```
adnanrahic, line 26: Setting The Time Threshold => Setting a Time Threshold

adnanrahic (changes requested):
Some tiny edits.

docs/experience/measurements.md
amir-hadzic: Yup, I linked in the first part but I'll add more "meat" there as Otis suggested.

mkdocs.yml
amir-hadzic: I was going with something short that is in sync with how we show it in the UI, but I think we can make it more descriptive in the docs since there's space there.

PR #328, opened 2021-02-04
Added a section about Transformations

otisg:
Just a bit of info we can point to from the new AR create screen.

### DIFF
--- a/docs/alerts/creating-logs-alerts.md
+++ b/docs/alerts/creating-logs-alerts.md
@@ -30,6 +30,18 @@ If you wanted to avoid using a specific threshold value and instead get notified
 
 ![image alt text](../images/alerts/image_6.png)
 
+# Transformation
+
+Transformations are used to modify metrics using mathematical expressions and functions. For example, imagine you want to get alerted when both `request.size` and `response.size` together exceed some threshold or become anomalous.  You would then transform them into a single dataseries by using an expression like this:
+
+```
+request.size + response.size
+```
+
+And then you would create an alert on this new data series.
+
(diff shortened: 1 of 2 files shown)


docs/alerts/creating-logs-alerts.md
```diff
@@ -30,6 +30,18 @@ If you wanted to avoid using a specific threshold value and instead get notified
...
+
+```
+request.size + response.size
+```
+
+And then you would create an alert on this new data series.
+
+Read about [Transformations](../dashboards/chart-builder/#transformation) to leadn more about transformations, functions, and expressions used to perform transformations.
+
 Notice that the chart changed to help you get a sense of what would constitute an anomalous value in your case (dots outside the gray *confidence interval*).
```
emir-arnautovic, line 45: Should this go before transformation section - it is referring to the image above?

docs/alerts/creating-metrics-alerts.md
```diff
@@ -34,6 +34,19 @@ If you don't want to use a specific threshold value and instead get notified whe
...
+
+Transformations are used to modify metrics using mathematical expressions and functions. For example, imagine you want to get alerted when both `os.cpu.user` and `os.cpu.system` together exceed some threshold or become anomalous.  You would then transform them into a single dataseries by using an expression like this:
+
+```
+os.cpu.user + os.cpu.system
+```
+
+And then you would create an alert on this new data series.
+
+Read about [Transformations](../dashboards/chart-builder/#transformation) to leadn more about transformations, functions, and expressions used to perform transformations.
```
emir-arnautovic, line 47: The same comments as for logs.

emir-arnautovic (approved):
Added section looks good, but I think it should be moved a but further down.

otisg:
Moved it/them down to the bottom.

PR #340, opened 2021-03-17
Redo containers log discovery

rabbitstack:
127 0 0 1_8000_logs_discovery_tailing-containers_


otisg:
+10 for that last paragraph.  Should we really leave that for the end?

I skimmed only the screenshot.  One thing to consider adding at the very top, in the current intro paragraph, is a link to the screen the doc page is about.  That way a person reading this and wanting to do this can easily open the relevant screen in SC and set it up.  Yes, they could look at the screenshot to see which icons are selected and then look at SC and make a few clicks to "get on the same page" (pun intended), but clicking would be simpler.  If we add the link we'll want US and EU versions.

rabbitstack:
You're the author of this paragraph :). It is explaining the motivation for defining labels upfront, thus its location in the docs.

I'll add the links.

adnanrahic (approved):

PR #352, opened 2021-05-31
Moved used images to images folder and deleted obsolete screenshots

emir-arnautovic:
localhost_8000_logs_pipelines_


docs/logs/pipelines.md
```diff
@@ -0,0 +1,88 @@
...
+![Upgrade Plan](../../images/logs/pipelines/upgrade-plan.png)
+
+Once app is on the right plan you can start configuring the pipeline.
+
+### Pipeline Builder
+Pipeline builder is opened as a full page and has basic structure as any other page.
+
+![Pipeline Builder](../../images/logs/pipelines/pipeline-builder-unsaved.png)
+
+On top we have header and subnav that can be used to pick documents for preview. The paga is split between resizable preview section and pipeline configuration section. Preview shows documents before and after processing. Pipeline configuration section has list of processors on the left and configuration details for a selected processor on the right.
```
wawanawna, line 22: `paga` is typo here it seems

wawanawna (approved):

otisg:
Yummmm.

A few comments:
* There are two screenshot that show the "You've made some changes to your pipeline" strip.  This may be more of a distraction in this context, except in the Saving Pipelines section where it's needed.
* Maybe it would be good to have a Processor section that defines/describes what a processor is and the concept of chaining N of them?  And only after that go into Processor configuration, which is more about how to actually use the UI around Processors.
* Not sure that we need to point out which plan this is available for.  For example, we don't say anything about that on https://sematext.com/docs/logs/archiving/ .  I'd remove that section from here.  There is enough material here without it.
* I see a mention of filtering.  Maybe this needs to be called out better, so it's more visible.  Maybe it needs to be in the Processor section I mentioned above.
* Would it make sense to have a list of all Processor and a sentence or two for each of them?
* Various little things, like app => App, pieline => pipeline, paga => page

docs/logs/pipelines.md
emir-arnautovic: fixed

emir-arnautovic:
> There are two screenshot that show the "You've made some changes to your pipeline" strip. This may be more of a distraction in this context, except in the Saving Pipelines section where it's needed.

Changed screenshot. The second one was copy/paste mistake - fixed it to reference correct image.

> Maybe it would be good to have a Processor section that defines/describes what a processor is and the concept of chaining N of them? And only after that go into Processor configuration, which is more about how to actually use the UI around Processors.

Change Processor configuration section to Processors and added paragraph about what processor is and how it is chained.

> Not sure that we need to point out which plan this is available for. For example, we don't say anything about that on https://sematext.com/docs/logs/archiving/ . I'd remove that section from here. There is enough material here without it.

Removed.

> I see a mention of filtering. Maybe this needs to be called out better, so it's more visible. Maybe it needs to be in the Processor section I mentioned above.

We have link from our UI to this section and that's the main reason why I added it. It is not about filtering processor but about filtering config for every processor - all processors will have filters.

> Would it make sense to have a list of all Processor and a sentence or two for each of them?

I was thinking about that, but did not want to do it at the moment since we do not have complete list of processors and even current one are likely to be changed/renamed.

> Various little things, like app => App, pieline => pipeline, paga => page

Fixed

docs/logs/pipelines.md
```diff
@@ -0,0 +1,82 @@
...
+In such cases, instead of selected processor configuration, there will be list of available processors. Click on any processor card will add processor with default values. Processors can also be added by clicking on "Add Processor" button.
+
+![Add Processor](../../images/logs/pipelines/add-processor.png)
+
+When clicked on the button, a modal with a searchable list of processors will appear to select processor to add.
+You can also add processor by duplicating some of existing processors.
+
+![Processor Actions](../../images/logs/pipelines/processor-actions.png)
+
+Duplicate action is under our standard action button in the list of processors. Note that order of processors matter since the output of one processor is the input for the next processor. You can reorder processor using drag handle left from the processor's name.
```
adnanrahic, line 38:
```suggestion
The duplicate action is below the standard action button in the list of processors. Note that **the order of processors matters** since the **output of one processor is the input for the next processor**. You can reorder processors using the drag handle to the left of the processor's name.
```

docs/logs/pipelines.md
```diff
@@ -0,0 +1,82 @@
...
+
+Selected processor configuration will be displayed on the right side. Every processor has at least name and description fields, but all processors will also have additional, processor specific configurations.
```
adnanrahic, line 40:
```suggestion
The selected processor configuration will be displayed on the right side. Every processor has at least a name and description field, but all processors will also have additional, processor-specific configurations.
```

docs/logs/pipelines.md
```diff
@@ -0,0 +1,82 @@
...
+
+##### Errors
+![Processor Errors](../../images/logs/pipelines/processor-errors.png)
+
+In case processor is not properly configured, nuber of errors will be displayed next to processor name and `Save Changes` button will be disabled.
```
adnanrahic, line 45:
```suggestion
In the case a processor is not configured correctly, the number of errors will be displayed next to the processor's name and the `Save Changes` button will be disabled.
```

docs/logs/pipelines.md
```diff
@@ -0,0 +1,82 @@
...
+
+##### Filters
+Almost all processors have optional filters. With filters, you can select only a subset of events that the processor should be applied to. Events that do not match filters will be ignored.
+
+![Processor Filters](../../images/logs/pipelines/processor-filters.png)
+
+It uses our standard filter component with difference that you can also enter field that is not a keyword or even not yet defined. Another important difference is that you can use wildcards for values. It supports standard wildcards `*`, matches any number of characters, and `?`, matches a single character.
```
adnanrahic, line 52:
```suggestion
Processors use the same filtering you're already used to in Sematext. The only difference is that you can also enter fields that are not keywords or even defined yet. Another important difference is that you can use wildcards for values. It supports standard wildcards `*`, matches any number of characters, and `?`, matches a single character.
```

docs/logs/pipelines.md
```diff
@@ -0,0 +1,82 @@
...
+This way you can e.g. drop events that have message field `*password*`.
+
+##### Sampling
+Events sampling is possible using Pipeline. It can be applied on all events or filters can be used to sample only some subset of events, e.g. only events from some service. Sample rate is expressed as percentage of events that should be stored. By default it is 100%, meaning no events will be dropped. Setting sampling to 10% means that only 1 in 10 events will be stored and 9 will be dropped.
```
adnanrahic, line 56:
```suggestion
You can do events sampling with Pipeline as well. It can be applied on all events, or by using filters you can sample only a subset of events. One example might be to filter by events from one specific service. The sample rate is expressed as a percentage of events that should be stored. By default, it is 100%, meaning no events will be dropped. Setting sampling to 10% means that only 1 in 10 events will be stored and 9 will be dropped.
```

docs/logs/pipelines.md
```diff
@@ -0,0 +1,82 @@
...
+Once input is manually edited, a warning will be displayed if you try to reload data.
+
+![Reload Warning](../../images/logs/pipelines/reload-warning.png)
+
+#### Saving Pipeline
+Changes made to pipelines are not automatically saved. Once you made some changes and all processors are configured, `Save Changes` button will be enabled and pipeline configuration can be saved.
+
+![Save Pipeline](../../images/logs/pipelines/save-pipeline.png)
+
+The new configuration will not be applied immediatelly. It can take up to 5 min before you can see effects on your changes. Make sure you adjust field types using [Field Editor](./fields/#field-editor).
```
adnanrahic, line 78:
```suggestion
The new configuration will not be applied immediately. It can take up to 5 minutes before your changes will take effect. Make sure you adjust field types using the [Field Editor](./fields/#field-editor).
```

adnanrahic (changes requested):
A few edits and we're done.

otisg:
@emir-arnautovic are you supposed to accept Adnan's changes here and then merge?

emir-arnautovic:
Yes, but I was not rushing that since we were blocked on releasing pipelines.

PR #354, opened 2021-07-05
SC-7715 Update Solr monitor integration docs to mention logs

radu-gheorghe:
Once we merge this PR, I guess we'll have to re-build/release the docs. Do you know who can do that?


otisg (changes requested):
app ==> App, but other than that, LGTM!

PR #363, opened 2021-08-11
Better describe the cross-origin isolation

gr0:
This PR better describes what need to be done to enable cross-origin isolation for a website and what are the things that need to be taken into consideration. It looks like this in the docs:

memory


gr0:
I merged that to avoid moving to next sprint. We can adjust "on the go" ;)

PR #364, opened 2021-08-11
SC-5762 custom metrics docs

bsmid:
Added docs that describe how to send custom metrics using influx line protocol.

### DIFF
--- a/docs/monitoring/custom-metrics.md
+++ b/docs/monitoring/custom-metrics.md
@@ -1,394 +1,108 @@
 title: Custom Metrics
-description: Sematext infrastructure and application monitoring service exposes APIs and provides libraries that let you send custom metrics and turn them into dashboards and real-time visualizations
+description: Sematext infrastructure and application monitoring service exposes APIs and provides endpoints that let you send custom metrics and turn them into dashboards and real-time visualizations
+    
+You can send custom metrics to any of your monitoring Apps. These metrics will be counted towards your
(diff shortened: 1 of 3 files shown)


docs/monitoring/custom-metrics.md
```diff
@@ -1,394 +1,108 @@
...
-
-``` bash
-curl -v -H 'Content-type: application/json' -d @datapoints.json http://spm-receiver.sematext.com/receiver/custom/receive.json?token=[spm app token]
+In this example we can see:
+- `appStats` - a `measurement` or a `namespace`, it describes the type of metrics in this particular line and it can be
+  anything that makes sense to you. It will be used as a prefix to all metrics in this line (so, full name of a metric
+  `user.requests.count` is actually `appStats.user.requests.count`)
+- `token` - a token of your Sematext monitoring App
+- `os.host` and `region` - two tags used to describe these metrics. You can append more tags by appending them after
+  `region=us` (tags have to be separated by a comma)
```
otisg, line 40: This region is unrelated to SC US vs. SC EU?  If so, maybe we could use another tag instead of 'region' to avoid potential confusion.

otisg (approved):
LGTM.  I'll tweak things a bit after the commit, will be simpler that way.

docs/monitoring/custom-metrics.md
bsmid: Yep, I'll change it to... something else.

PR #365, opened 2021-08-18
Wording updates

otisg:


otisg:
> Once the affected resources are identified you can use the **crossorigin** attribute in the HTML tags that are known to be server with the Cross-Origin-Resource-Sharing, simply CORS. You can for example do the following:

Couldn't figure out this sentence.  That "HTML tags that are known to be server with the Cross-Origin-Resource-Sharing" part sounds like some word is missing.

otisg:
What about that comment/question above, @gr0 ?

gr0:
Ahhh... I made the change manually to the master, should have put it here. But for reference if you wish to adjust: https://github.com/sematext/docs/commit/637c451697a838ee4a8629dd78b7465832838e77

PR #366, opened 2021-09-08
document newly added Elasticsearch metrics

radu-gheorghe:
Added the new metrics from https://github.com/sematext/sematext-agent-integrations/pull/46


emir-arnautovic (approved):
Looks good!

PR #367, opened 2021-09-08
SC-11882: Docker OSX STA instructions

rabbitstack:
Docker OSX instructions for STA in addition to clarification about collected OS metrics.


otisg (approved):
LGTM

PR #368, opened 2021-09-28
SC-12152 Add Google Chat Docs

PabloB94:


docs/integration/alerts-googlechat-integration.md
```diff
@@ -0,0 +1,50 @@
...
+
+**3.** Give your space a name and click **Create**:
+
+<img class="content-modal-image" alt="Create Google Chat Integration - Create Space" src="../../images/integrations/create-googlechat-integration-create.png" title="Create Google Chat Integration - Create Space">
+
+**4.** Open the drop down menu with your space's name and select **Manage webhooks**:
+
+<img class="content-modal-image" alt="Create Google Chat Integration - Manage Webhooks" src="../../images/integrations/create-googlechat-integration-manage-webhooks.png" title="Create Google Chat Integration - Manage Webhooks">
+
+**5.** Give your webhook a name (and optionally an avatar) and click **Save**:
```
otisg, line 24: Is webhook the right term to use here?  Is that what they use?  Asking because we use "Webhook" for a generic type of "notify by calling some URL" sort of notification hook.

docs/integration/alerts-googlechat-integration.md
```diff
@@ -0,0 +1,50 @@
...
+
+<img class="content-modal-image" alt="Create Google Chat Integration - Copy Webhook" src="../../images/integrations/create-googlechat-integration-copy-webhooks.png" title="Create Google Chat Integration - Copy Webhook">
+
+With the **Incoming Webhook** configured we can now use the **Webhook URL** to configure the Google Chat notification hook in Sematext.
+
+## In Sematext
+
+**1.** Navigate to [Notification Hooks](https://apps.sematext.com/ui/hooks/create) (in [EU](https://apps.eu.sematext.com/ui/hooks/create)) and select Google Chat card to create a new Google Chat notification hook.
+
+<img class="content-modal-image" alt="Create Google Chat Integration - New Webhook" src="../../images/integrations/create-googlechat-integration-new-hook.png" title="Create Google Chat Integration - New Webhook">
```
otisg, line 38: New Webhook ==> New Notification Hook

docs/integration/alerts-googlechat-integration.md
```diff
@@ -0,0 +1,50 @@
...
+
+**2.** Give your webhook a name and add your Google Chat **Webhook URL**.
```
otisg, line 40: webhook ==> Notification Hook
(and the same sort of changes further down)

otisg (approved):
Just minor terminology change needed, otherwise LGTM. (note: I didn't actually see how this looks in terms of formatting, I just reviewed the copy)

docs/integration/alerts-googlechat-integration.md
PabloB94: The dialog in Google chat is called "Incoming Webhooks" and the field just says "Name", so I believe it's the right term to use here simply to match the UI. Do you think we should change it?

docs/integration/alerts-googlechat-integration.md
PabloB94: Here on line 40, since it comes from Google and they call it Webhook, I'd keep it. 
For everything on our side, I'm making the change. What do you think?

otisg:
Deployed, see https://sematext.com/docs/integration/alerts-googlechat-integration/
But shouldn't we also add it to the grid at https://sematext.com/docs/alerts/alert-notifications/ ?

PabloB94:
I'll get on that asap. Signl4 is missing from the grid, should I add it too?
The icons in the docs and the app are different, should they be unified?

otisg:
Yes to both, nice catch. @harisovcinaa could tel lus which are the right icons to use.

PR #383, opened 2022-02-15
Added FAQ for Apps on different envs

radu-gheorghe:
I've added this to the general FAQ, since it applies to both logs and monitoring.


otisg:
Merci!

radu-gheorghe:
De rien! Merci for deploying and merging and removing the branch!

PR #390, opened 2022-04-25
Update using-the-api.md

michaelhigham:
Made amendments based on feedback from Otis


michaelhigham:
Needs further amendments

PR #395, opened 2022-05-23
Update http-monitor.md

michaelhigham:
Added Authorization section

### DIFF
--- a/docs/synthetics/faq.md
+++ b/docs/synthetics/faq.md
@@ -30,7 +30,7 @@ If you do not want to run private agents yourself, you can also configure the fi
 No. The HTTP monitor does not support OAuth authentication before sending the actual request. HTTP monitor supports header-based authentication. 
 
 ### Can I monitor API endpoints protected by Basic authentication?
-Yes, the HTTP monitor supports header-based authentication. Custom HTTP request headers can be specified when creating a monitor.  For Basic authentication add the `Authorization: Basic XYZ` header where XYZ is the Base64 encoding of the username:password string.
+Yes, please see [authorization](./http-monitor/#authorization).
 
 ### Can I stop Synthetics Alerts during scheduled maintenance?
 Yes. You can temporarily disable the notifications from Edit Configuration -> Configure Alerts -> Notifications. Once the maintenance window is over, you can enable the notifications.
--- a/docs/synthetics/faq.md
+++ b/docs/synthetics/faq.md
@@ -30,7 +30,7 @@ If you do not want to run private agents yourself, you can also configure the fi
 No. The HTTP monitor does not support OAuth authentication before sending the actual request. HTTP monitor supports header-based authentication. 
 
 ### Can I monitor API endpoints protected by Basic authentication?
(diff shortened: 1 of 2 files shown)


docs/synthetics/faq.md
```diff
@@ -30,7 +30,7 @@ If you do not want to run private agents yourself, you can also configure the fi
 No. The HTTP monitor does not support OAuth authentication before sending the actual request. HTTP monitor supports header-based authentication. 
 
 ### Can I monitor API endpoints protected by Basic authentication?
-Yes, the HTTP monitor supports header-based authentication. Custom HTTP request headers can be specified when creating a monitor.  For Basic authentication add the `Authorization: Basic XYZ` header where XYZ is the Base64 encoding of the username:password string.
```
otisg, line 33: A person who may be reading this and is using Browser monitor might be mislead because they will not see "HTTP monitor" here.

docs/synthetics/http-monitor.md
```diff
@@ -1,24 +1,24 @@
...
+title: HTTP Monitor  
+description: HTTP monitor can monitor any HTTP endpoint like API, Web URL, etc.
 
 The HTTP monitor sends a single HTTP request to the specified URL with configured request settings. It records the response data like headers, body, error if any and the various timings.
 
 ## Configuration
 
 ### General
 
-* **Name** - Name of the monitor. Max length is 255 characters.
```
otisg, line 10: Looks like some characters were added/inserted here and in several places below.Not sure if accidentally or on purpose or if that changes how things are rendered or not.  JFYI.

docs/synthetics/http-monitor.md
```diff
@@ -27,35 +27,57 @@ Max length of the header, query param, and cookie name and value are 512 and 102
...
+
+| Name | Value |
+| --- | --- |
+| `Authorization` | `Basic dXNlcm5hbWU6cGFzc3dvcmQ=` |
+
+
+It is possible to get the **Base64** encoded value of your `username` and `password` by using the following command in a terminal, you should replace `username` with your username and `password` with your password respectively:
+```
+echo -n "username:password" | base64
+```
```
otisg, line 54: In case somebody is reading but doesn't have terminal handy for some reason (mobile device?) maybe we could also point to some online Base64 encoders like https://www.base64encode.org/

otisg (approved):
LGTM. Minor suggestions.

docs/synthetics/http-monitor.md
michaelhigham: Yes I don't know how this happened, I will remove them

docs/synthetics/faq.md
michaelhigham: OK, I will re-add a short sentence highlighting this.

docs/synthetics/http-monitor.md
michaelhigham: I was in two minds about this, as linking to external websites can sometimes be a bad idea in relation to SEO. I will add in the link.

PR #405, opened 2022-07-13
Added Fluen Bit Instructions

michaelhigham:
Added Fluent Bit instructions for configuring Fluent Bit with the Elasticsearch API


otisg:
Could also add Fluentbit to https://github.com/sematext/docs/blob/master/docs/logs/sending-log-events.md

PR #414, opened 2022-08-17
SC-13436 Update Synthetics documentation

hsalsem:
Turns out these haven't been updated in a while, so I've went through all of them and fixed a bunch of stuff. Here are some of the things included in this PR:
- Fixed typos
- Reworded some phrases
- Improved styling on some parts (converted some info to tables, made things bold/italic to emphasize things, made formatting consistent across pages etc.)
- Changed capitalization of certain phrases to ensure consistency across the docs (such as having "Browser monitor" everywhere instead of alternating between that and "browser monitor")
- Added some new text to document the new features or changes in existing ones
- Improved some screenshots (cropped out empty space or edited the images to have less of it and be more legible)
- Replaced some screenshots due to outdated UI
- Added some new screenshots due to new features
- Updated Puppeteer/Chrome versions we use

The `user-journey-scripts.md` file still has many instances of "Sematext" when referring to Sematext Cloud (the app, not the company). I haven't changed this because it's mentioned many times, but I can if you feel that I should.


docs/synthetics/faq.md
```diff
@@ -5,13 +5,13 @@ description: FAQ about Sematext Synthetics, a synthetics monitoring solution for
 
 ### What should I do if I can't find the answer to my question in this FAQ?
 Check the [general FAQ](/faq) for questions that are not strictly
-about Sematext Synthetics.  If you can't find the answer to your
-question please email <support@sematext.com> or use our live chat.
+about Sematext Synthetics. If you can't find the answer to your
+question please email <support@sematext.com> or use our live chat in the bottom right corner of the page.
```
otisg, line 9: On www yes, but not in SC, which is probably why we didn't try to be so specific.

docs/synthetics/status-pages.md
```diff
@@ -5,51 +5,56 @@ Using Sematext Synthetics you can create public and private status pages to shar
 
 * Communicate the status of your services to your customers
 * Share status updates with your customers during planned maintenance or outages
-* Create a dashboard of service status to share across internal teams
+* Create a service status dashboard to share across internal teams
```
otisg, line 8: dashboard ==> page?

otisg (approved):
LGreatTM.  Just two minor suggestions.

docs/synthetics/faq.md
hsalsem: Yes, but here we're telling people to check the FAQ and contact support if their questions aren't answered there. FAQ also has the live chat down in the corner, which is why I figured that people are probably going to try and contact support there instead of going to SC. I can change it back to be less specific if you'd like, but that was the reasoning behind the change.

docs/synthetics/status-pages.md
hsalsem: I wasn't sure about this one either, but left it as dashboard since the status pages give an overview of multiple monitors, incidents, and have charts on them, so they are kind-of sort-of dashboards. Dashboard also sounds fancier in these starting bullet points as a selling point, that could also be why it's there  ¯\_(ツ)_/¯

docs/synthetics/faq.md
otisg: Aha, OK, makes sense.

docs/synthetics/status-pages.md
otisg: Sure, let's let status dashboard live then.

PR #415, opened 2022-08-31
SC-12166 Document Synthetics custom metrics

hsalsem:
Documented how to define custom metrics, how to chart them and create alerts for them. Also took into account the potential new user whose use case is grabbing specific metrics from a JSON and created an example for that scenario too. I'll add that example to our UI as well once this is reviewed.

Perhaps unrelated to the ticket, but I've also documented how to create custom monitor alerts for specific values in the response body JSON. These refer to the in-monitor alerts, not the separate ones we define on custom metrics, so they're simpler to create and they cause the monitor to fail if they're not satisfied.


docs/synthetics/http-monitor.md
```diff
@@ -74,6 +74,63 @@ By default, the UI adds the below conditions while creating an HTTP monitor. You
 * **Response Code** *equals* `200`
 * **Response Time** metric *less than* **20000 ms**
 
+
+### Configuring Custom Alerts based on JSON response
+In addition to our preconfigured alert conditions, you can also create your own based on the JSON returned in the response body. As mentioned above, this can be done using the **Response Body JSON** type, where [JSONPath](https://github.com/json-path/JsonPath) can be used to extract a specific value and compare it against the expected value.
```
otisg, line 79: When I was writing Lucene in Action the editors regularly had beef with "position references" like "As mentioned above".  OK, that was a printed book, so above/below didn't make a sense in a physical book.  Does "As mentioned above" add value here?

docs/synthetics/http-monitor.md
```diff
@@ -74,6 +74,63 @@ By default, the UI adds the below conditions while creating an HTTP monitor. You
...
+  ],
+  "serviceStatusCode":"2"
+}
+```
+
+With this JSON structure in mind, you can define a **Response Body JSON** alert condition like in the image below.
+
+![Custom Alert Condition](../images/synthetics/custom-alert-condition.png)
+
+The field is defined as `$.results[:].status`, meaning that all the `status` fields inside of `results` will be checked, and the monitor run will fail if any of them are equal to `Error`. Here's an example of a run failing upon one of these services reporting an error.
```
otisg, line 129: I am not be seeing this in context in this PR, but it's a little confusing to go from talking about custom *metrics* (which I imagine as numerical values) to an example about looking for strings like `Error` in a response.  Ah, no, no, you didn't say "metrics" specifically above.  Ignore this.  But it will be interesting to see if we document how to  extract and what to do with both numerical values and string values.

docs/synthetics/metrics.md
```diff
@@ -58,4 +58,85 @@ Browser monitor collects the below metrics for every resource loaded during the
...
\ No newline at end of file
+You can script the Browser monitor to collect custom metrics that are specific to your website or use case. For example, you might need to measure the time it takes to display auto-suggestions on your website or collect & monitor the value from an element on your webpage. You can use the `context.setMetric(name, value)` method in the Browser monitor script in order to define a custom metric. All metrics defined in monitors within one Synthetics App should have unique names and will be automatically pre-pended with a `synthetics.browser.custom.` prefix.
```
otisg, line 61: pre-pended => prefixed

docs/synthetics/metrics.md
```diff
@@ -58,4 +58,85 @@ Browser monitor collects the below metrics for every resource loaded during the
...
+
+Here's an example of a Browser monitor script which defines two custom metrics: the used JavaScript heap size and how long the combined duration of JavaScript execution is. For your convenience, this example is also listed as **Using Custom Metrics** under the **Browse Examples** section while creating the Browser monitor.
```
otisg, line 63: Is "Browser monitor script" referring to UJS?

otisg (changes requested):
Looks good, minor changes suggested.  However, I think we are missing information about custom metrics w/ HTTP monitors.  Maybe you planned on adding that later?

Also, both examples here are about performance - JS heap, CPU, etc.  But we also want to show this could be used for *any* numerical value.  It doesn't have to do anything with performance.  For example, one could be hitting some page or API with temperature in Honolulu and extracting that for charting.  I think some examples like that would help users understand he possibilities...

docs/synthetics/http-monitor.md
hsalsem: Just stating that we already mentioned this a bit earlier in case some user is thinking "hey you already covered this", while also having this section contain all the information it needs on its own for people who just skip to a particular section.

docs/synthetics/http-monitor.md
hsalsem: The approach should be identical as people will use JSONPath in the same fashion for both strings and numerical values

docs/synthetics/metrics.md
hsalsem: Yeah, want me to change it to UJS?

docs/synthetics/http-monitor.md
```diff
@@ -64,7 +64,7 @@ Conditions can be configured on the response fields and metrics. The conditions
 * **Metric** - Used to make sure the metrics are within the expected range.
 * **Response Header** - Can be used to compare the value of a specific header field against the expected value.
 * **Response Body** - Compare the response body against the specified value using an operator.
-* **Response Body JSON** - If the response body is JSON, [JSONPath](https://github.com/json-path/JsonPath) can be used to extract a specific value and compare against the expected value.
+* **Response Body JSON** - If the response body is JSON, [JSONPath](https://github.com/json-path/JsonPath) can be used to extract a specific value and compare against the expected value. More information on how to set these conditions up is provided in the [next subsection](#configuring-custom-alerts-based-on-json-response).
```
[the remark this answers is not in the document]
hsalsem: This would link to a section on this page, just a few lines below. This also has nothing to do with custom metrics, but for creating monitor alert conditions based on the response.

docs/synthetics/metrics.md
otisg: We should use consistent terminology in the docs and that terminology needs to match the terminology in the UI.

hsalsem:
I've changed the JSON example from performance to finance - the use case being getting currency exchange rates to chart them over time and alert on them if they reach a certain value. A bonus side effect of this is that this uses a free existing API, so that users can interact with it when they're testing their scripts and get a better idea of how to work with Custom metrics. The reason why I've set it to CPU/RAM before was for the ML customer (though they'd probably want GPU instead of CPU, but I wanted it to be general as well). Will change to temperature in Honolulu if we go there for one of our teambuilding trips 🌴

From my understanding we don't support custom metrics for HTTP monitors, only for Browser monitors by explicitly setting them in Chrome with the `context.setMetric()` API, which is why we have this subsection under **Browser Monitor** instead of general metrics.

otisg (approved):

PR #424, opened 2022-10-04
Synthetics API - Create Monitor Alert priority

michaelhigham:
Added alert priority key and values to Synthetics API create monitor.


HSSalman:
Looks good, thanks for this Michael 👍

PR #426, opened 2022-10-10
Add Alert Priority information to Synthetics docs

HSSalman:
Reworded a few things, moved the `Alerts` page from Synthetics to Alerts to be in line with what we did for the other apps, and added information about alert rule priority to the appropriate sections.


fulyauluturk:
LGTM

PR #427, opened 2022-10-13
Content and screenshots updated

fulyauluturk:
Some of the content were outdated and not aligned with how SC behaves right now. So I updated the content. Removed component descriptions like time picker, live tail. These are explained in detail in Reports & Components section. This page is more about App actions, general look, integrations and side navigations.


cpipilas (approved):
LGTM, I would only double check Reports vs reports naming.

PR #429, opened 2022-10-13
Updated out dated content.

fulyauluturk:
Updated the out dated content to align with how SC behaves right now. Like I did for Logs/settings I removed the time picker and live tail component description section. These will be added to Reports & Components page along with other things such as connect Apps etc.


cpipilas (approved):
LGTM

PR #444, opened 2022-11-17
Events fields and tags added to common schema

fulyauluturk:
I will remove https://sematext.com/docs/events/fields/ page after review to avoid duplicate content. And add a nav link at the bottom https://sematext.com/docs/events/adding/ page saying "See common Sematext Common Schema for recommended event fields."

### DIFF
--- a/docs/tags/common-schema.md
+++ b/docs/tags/common-schema.md
@@ -183,3 +183,78 @@ Below are tags that are reserved for future use:
 | service.type | Service type e.g. `hadoop` |
 | span.id | Building block of a trace in distributed tracing |
 | trace.id | Building block of a trace in distributed tracing |
+
+## Events Tags
+
+### Fields
+
+When you create an event you can send a ```JSON``` document which consists of multiple
+fields. Each field can contain event main information or metadata. Even though there is
+no strict format of such a ```JSON``` document we recommend to some of fields. An event
+can contain the following set of fields, most of which are optional:
+
+Field Name | Field Type | Required | Notes 
+-----------|------------|----------|-------
+```timestamp``` | date       | no       | Time when event happened (if not specified,current time will be assumed). The format is [dateOptionalTime](https://joda-time.sourceforge.netapi-release/org/joda/time/format/ISODateTimeFormat.html#dateOptionalTimeParser) e.g.: ```2014-02-17T21:37:04+0100``` or```  2014-02-17T14:15:01.534471+02:00```.
+```os.host``` | string       | no       | Name of the host where the event has occurred.
+```type``` | string | yes | Event type which could be e.g. ```alert```, ```deployment```, etc. Events are later grouped in timeline based on event type which significantly improves visibility.
+```message``` | string       | yes       | Short description of event, e.g. ```Elasticsearch node03 onhost somehost06 restarted```. This is a default search field in Sematext UI, so it is good to keep it concise, but search-friendly. Data in this field can be stored in Markdown format to make your messages more pretty and easier to read. For more details [see](#markdown-in-events).
+```title``` | string       | no       | Event title, can be used as a short label for event, e.g. ```Elasticsearch restart```.
+```tags``` | string array     | no       | Multivalued field. Each tag should be specified as aseparate array element e.g., <br> ```"tags":[ "elasticsearch", "restart", "emergency fix"]```
+```severity``` | string | no | A single-valued field which says what kind of an event it is. It should have such values as ```error```,  ```info``` or ```warning``` and let's you easily navigate through important and less important events.
+```creator``` | string       | no       | Person, application, or component that created an event. E.g. ```John Smith```, ```Elasticsearch```, ```Some Batch Job```
+```data``` | string       | no       | Additional event data. It can be anything you may find useful to have along inside of event object. E.g., it could be  stacktrace in case of ```app_error``` event, base64 encoded content of file, etc.
+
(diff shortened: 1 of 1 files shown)


otisg:
@fulyauluturk note that on https://sematext.com/docs/events/fields/ there is a bit of "troubleshooting" info, which we may want to preserve (move to events/faq?) and also a bit of info about some examples of "event types". Maybe this is a sort of "Event Best Practices" sort of doc/content material.  There is also a mention an example of usage of markdown in events.  Maybe we could move that to https://sematext.com/docs/events/adding/#adding-events-through-ui prior to removing https://sematext.com/docs/events/fields/  ?

docs/tags/common-schema.md
```diff
@@ -183,3 +183,78 @@ Below are tags that are reserved for future use:
 | service.type | Service type e.g. `hadoop` |
 | span.id | Building block of a trace in distributed tracing |
 | trace.id | Building block of a trace in distributed tracing |
+
+## Events Tags
+
+### Fields
+
+When you create an event you can send a ```JSON``` document which consists of multiple
```
otisg, line 191: @fulyauluturk note how we use both the word "tags" and "fields".  Maybe we need to stick with one?

docs/tags/common-schema.md
```diff
@@ -183,3 +183,78 @@ Below are tags that are reserved for future use:
...
+fields. Each field can contain event main information or metadata. Even though there is
+no strict format of such a ```JSON``` document we recommend to some of fields. An event
+can contain the following set of fields, most of which are optional:
```
otisg, line 194: @fulyauluturk most are optional or all are optional?
What happens if timestamp is not provided?

docs/tags/common-schema.md
```diff
@@ -183,3 +183,78 @@ Below are tags that are reserved for future use:
...
+
+Field Name | Field Type | Required | Notes 
+-----------|------------|----------|-------
+```timestamp``` | date       | no       | Time when event happened (if not specified,current time will be assumed). The format is [dateOptionalTime](https://joda-time.sourceforge.netapi-release/org/joda/time/format/ISODateTimeFormat.html#dateOptionalTimeParser) e.g.: ```2014-02-17T21:37:04+0100``` or```  2014-02-17T14:15:01.534471+02:00```.
+```os.host``` | string       | no       | Name of the host where the event has occurred.
+```type``` | string | yes | Event type which could be e.g. ```alert```, ```deployment```, etc. Events are later grouped in timeline based on event type which significantly improves visibility.
+```message``` | string       | yes       | Short description of event, e.g. ```Elasticsearch node03 onhost somehost06 restarted```. This is a default search field in Sematext UI, so it is good to keep it concise, but search-friendly. Data in this field can be stored in Markdown format to make your messages more pretty and easier to read. For more details [see](#markdown-in-events).
```
otisg, line 201: onhost => on hst

docs/tags/common-schema.md
```diff
@@ -183,3 +183,78 @@ Below are tags that are reserved for future use:
...
+```data``` | string       | no       | Additional event data. It can be anything you may find useful to have along inside of event object. E.g., it could be  stacktrace in case of ```app_error``` event, base64 encoded content of file, etc.
+
+When you ship events to Sematext Cloud, we recommend you to use common fields listed in this page. This way you can easily correlate between your events, metrics and logs. For example, you see CPU usage spikes or start seeing more error logs and you want to investigate it further. You can use [Split Screen](../guide/split-screen) and load the Events in the right half and see if there was any deployment type of event that might be the cause of that spike. Or search for events that are shipped from a specific host that started using more resources than expected. This way you can pinpoint the source of the problems easily.
+
+### Types
+
+[Sematext Cloud](https://sematext.com/cloud/) helps you manage your events better and
+distinguish between different kinds of events when you provide event type as a ```JSON```
```
otisg, line 213: Events are always sent as JSON, so isn't saying "JSON field"  a bit redundant?  IIUC the key is to send "event type" and call the field "type".

docs/tags/common-schema.md
```diff
@@ -183,3 +183,78 @@ Below are tags that are reserved for future use:
...
+```creator``` | string       | no       | Person, application, or component that created an event. E.g. ```John Smith```, ```Elasticsearch```, ```Some Batch Job```
+```data``` | string       | no       | Additional event data. It can be anything you may find useful to have along inside of event object. E.g., it could be  stacktrace in case of ```app_error``` event, base64 encoded content of file, etc.
+
+When you ship events to Sematext Cloud, we recommend you to use common fields listed in this page. This way you can easily correlate between your events, metrics and logs. For example, you see CPU usage spikes or start seeing more error logs and you want to investigate it further. You can use [Split Screen](../guide/split-screen) and load the Events in the right half and see if there was any deployment type of event that might be the cause of that spike. Or search for events that are shipped from a specific host that started using more resources than expected. This way you can pinpoint the source of the problems easily.
```
otisg, line 208: we recommend you to use ==> we recommend using the
For example, you ==> For example, when you
in the right half ==> in the right half of the screen

docs/tags/common-schema.md
fulyauluturk: I think when you try to send an event without timestamp it fails.

otisg:
@fulyauluturk Looked at https://sematext.com/docs/tags/common-schema/ but I don't see it.  I won't redeploy the docs in case you had a reason not to deploy them.

fulyauluturk:
I don't know how to deploy things tbh. I ask someone from the engineering team to do it usually. For this one I didn't because I thought it is Friday and naturally someone is going to deploy something. Didn't know we need to deploy docs specifically. I'll ask someone to do it on Monday.

PR #454, opened 2022-12-13
Improved wording

otisg:
Partial update.

### DIFF
--- a/docs/guide/integrations-guide.md
+++ b/docs/guide/integrations-guide.md
@@ -1,33 +1,25 @@
 title: Sematext Integrations Guide
-description: Sematext Cloud provides over 40 built-in integrations used to monitor across all your systems, Apps, and services.
+description: Sematext Cloud has 100+ built-in integrations for monitoring services, infrastructure, websites, APIs, etc.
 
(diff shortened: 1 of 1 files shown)


otisg:
Do you think we need stuff below **App Table View** in this page?

fulyauluturk:
I mean it is not 100% necessary. The page is pretty self explanatory users can discover these things easily without docs. What's more important is to explain some of these items in the menu which we already do in other solution type specific pages. So yes we can remove it.

fulyauluturk (approved):
I mean it is not 100% necessary. The page is pretty self explanatory users can discover these things easily without docs. What's more important is to explain some of these items in the menu which we already do in other solution type specific pages. So yes we can remove it.

cpipilas:
> Do you think we need stuff below **App Table View** in this page? Or could we remove it?

I would agree on removing this, up to that point we have all the basic information needed for Integrations, we can link the page with https://sematext.com/docs/integration/ for more information.

I would also suggest to have a similar approach (a table with tiles for each integration) on listing Integrations as we have at https://sematext.com/docs/integration/. Not only for Monitoring but for Logs and Notifications as well.

PR #461, opened 2023-02-01
Video content updates

fulyauluturk:


docs/logs/reports-and-components.md
```diff
@@ -27,7 +27,7 @@ Each alert query is visible in the Alert Rules list.
 
 ![Sematext Cloud Logs Guide - Connect Apps](../images/guide/logs/connect-apps_2.png) 
 
-- **Split View** - Split Screen is a product-wide correlation and data-pivoting feature. With Split Screen you can compare any two reports or dashboards, even the same report or dashboard but with different filters and groups in the two different screens.
+- **Split View** - [Split Screen](../guide/split-screen) is a product-wide correlation and data-pivoting feature. With [Split Screen](../guide/split-screen) you can compare any two reports or dashboards, even the same report or dashboard but with different filters and groups in the two different screens.
```
cpipilas, line 30: I believe we should stick with Split Screen and not use the term Split View. I think Split View is only referenced in Logs.

cpipilas (commented):
LGTM, only one comment about Split View replacement.

PR #473, opened 2023-03-08
cpi SC-15688 add OpenSearch in Service Autodiscovery

cpipilas:


otisg:
No need for a reviewer here.

PR #505, opened 2023-07-11
Google Group settings added

fulyauluturk:
Configure group setting to allow posts from external emails section added.


otisg:
LGTM

PR #506, opened 2023-07-17
Switch to team account section added.

fulyauluturk:


docs/team/index.md
```diff
@@ -79,10 +79,12 @@ If you've set up Apps, Alert Rules, etc. in your own account and would like to c
...
 If there is anything in your account that you do not wish to have in the Team Account it needs to be deleted from your account first.  
-After the conversion is done, your account with the old email login will be completely empty but you will have access to the Team Account and will be able to switch into it.
 
 ![Convert To Team Account Form](../images/team/convert_to_team_account.png)
 
+After the conversion is done, your account with the old email login will be completely empty and you will be logged out. In order to switch to the team account, login with your individual account email and then switch into the team account.
```
mbonaci, line 85: I wouldn't add it here, at the end of account conversion, but at the bottom of the "account members" page: https://sematext.com/docs/team/account-members/  

This should be a part of the general team account docs, while the "Account conversion" is like a fix we provide for those who initially failed to setup team account correctly.  

BTW you haven't pushed the `switch-to-team-account.png` image.

fulyauluturk: The image was uploaded directly into master. 

> I wouldn't add it here, at the end of account conversion, but at the bottom of the "account members" page

I think it is better to mention it in both places. I'll add there too. Because here it makes sense also. After one converts to team account they might not now they need to login and then switch to the account.

fulyauluturk: @mbonaci added to [account-members page](https://sematext.com/docs/team/account-members/)

PR #512, opened 2023-08-02
SC-16965 update kubernetes metrics

cpipilas:


yago-123:
LGTM, can't aprove the PR because the button does not show up for me

PR #522, opened 2023-08-30
Add documentation for the TLS paths

yago-123:


yago-123:
@cpipilas anything that may be worth adding?

yago-123:
> > @cpipilas anything that may be worth adding?
> 
> I'm missing some usage examples, specially for certs it would be really helpful.

I'm not sure what else would you expect aside from the example that I set. Is relatively straightforward this operation... Could you give any idea?

cpipilas (approved):

docs/integration/kubernetes.md
```diff
@@ -823,3 +823,22 @@ Also, please make sure that your agent is [up to date](https://sematext.com/docs
 
 ### Why we need hostNetwork access and how to turn it off if desired
 Check out [our page about hostnetwork](https://sematext.com/docs/agents/sematext-agent/kubernetes/hostnetwork/).
+
+### Setting TLS Certificate Paths
```
cpipilas, line 827: @yago-123 for some reason I missed that section. We are good with this.

cpipilas:
> > > @cpipilas anything that may be worth adding?
> > 
> > 
> > I'm missing some usage examples, specially for certs it would be really helpful.
> 
> I'm not sure what else would you expect aside from the example that I set. Is relatively straightforward this operation... Could you give any idea?



> > > @cpipilas anything that may be worth adding?
> > 
> > 
> > I'm missing some usage examples, specially for certs it would be really helpful.
> 
> I'm not sure what else would you expect aside from the example that I set. Is relatively straightforward this operation... Could you give any idea?

We are good: https://github.com/sematext/docs/pull/522#discussion_r1312915375

docs/integration/kubernetes.md
yago-123: OK, asking for review for the final boss then

docs/integration/kubernetes.md
```diff
@@ -823,3 +823,22 @@ Also, please make sure that your agent is [up to date](https://sematext.com/docs
...
+While monitoring Kubernetes master components, the Sematext agent will automatically retrieve certificates from the host machine.
```
otisg, line 828: Sematext agent ==> Sematext Agent (we treat it as a name elsewhere, I think?)

otisg (approved):
LGTM

docs/integration/kubernetes.md
yago-123: Modified, also checked if there were more cases like this, didn't found places in which we don't use Sematext Agent

PR #514, opened 2023-08-08
SC-12905 improve and add metric descriptions

yagogou:

PR #523, opened 2023-08-31
Sc 13031 modify columns to fit unit and type

yago-123:


yago-123:
@cpipilas I'm not very convinced with the units named `number`, any alternative?

cpipilas:
> @cpipilas I'm not very convinced with the units named `number`, any alternative?

Not sure I follow, can you clarify? Last discussion we had was --> https://github.com/sematext/docs/pull/514#discussion_r1301351873

cpipilas:
I would prefer to leave `Unit` empty in all the cases that you have added `number`.

yago-123:
> I would prefer to leave `Unit` empty in all the cases that you have added `number`.

I agree, just deleted

cpipilas (approved):
LGTM

PR #534, opened 2023-10-19
Harisovcinaa color guidelines

harisovcinaa:
Hey folks, could you take a look at the guidelines, thanks!


fulyauluturk:
@harisovcinaa the content LGTM. Although I forgot to show you one thing. Whenever you add a new page, you should also add it to mkdocs.yml to make it visible in the website. I just added that change in your PR, this is JFYI.
And I also didn't see where you use the videos you've uploaded.

otisg:
Might be better to put the videos in the CDN

PR #542, opened 2023-11-01
SC-16926 update kubernetes integration

cpipilas:
Updates in Kubernetes integration:

- Added new Workloads group with important metrics for Pods, Deployments, DeamonSets, StatefulSets, CronJobs, Jobs
- Added metrics for DaemonSets and StatefulSets 
- Updated images
- Formatting updates
- Moved all metrics in separate page, kept only important metrics


docs/agents/sematext-agent/kubernetes/metrics.md
```diff
@@ -1,42 +1,763 @@
...
+
+
+### CronJob Metrics
+
+| Metric Name                     | Type    | Unit      | Description             |
+|---------------------------------|---------|-----------|-------------------------|
+| kubernetes.cronjob.name         | tag     |           | Cronjob name            |
+| kubernetes.cronjob.cluster.name | tag     |           | cluster name            |
+| kubernetes.cronjob.namespace    | tag     |           | namespace               |
+| kubernetes.cronjob.scheduled    | tag     |           | schedule setup  [ * 1 * * ] |
```
yago-123, line 68: We don't put tags in this document. We should include all tags for all the other metrics (which would be useful) or remove the tags from here

docs/integration/kubernetes.md
```diff
@@ -3,11 +3,12 @@ description: Kubernetes is a portable, extensible, open-source platform for mana
 
 Kubernetes is a portable, extensible, open-source platform for managing containerized workloads and services, that facilitates both declarative configuration and automation. To start monitoring Kubernetes with Sematext, you only need to install a tiny agent that adds basically no CPU or memory overhead.
 
+
 ## Monitoring Kubernetes with Sematext
 
 Sematext Monitoring will provide you with detailed insights into your cluster's control plane components and their health, performance metrics, and resource counts, among other important metrics. Speaking of metrics, you can check out [this page](https://sematext.com/docs/agents/sematext-agent/kubernetes/metrics/) for a summarized list of key metrics that you can track using Sematext. It also includes a short explanation for each metric.
 
-![alt_text](https://sematext.com/wp-content/uploads/2020/03/Kubernetes-Pod-Overview.png "Sematext Monitoring")
+![alt_text](https://sematext.com/wp-content/uploads/2023/10/k8s-overview.png "Sematext Kubernetes Monitoring")
```
yago-123, line 11: maybe instead of k8s we can use kubernetes? Not sure if that affects SEO or something like that, but seems more "aesthetic"

docs/agents/sematext-agent/kubernetes/metrics.md
cpipilas: Correct!

docs/integration/kubernetes.md
cpipilas: I see. This is the file name, referenced by other docs as well so let's keep it as is. In general we use "Kubernetes".

yago-123 (approved):
LGTM

PR #551, opened 2023-11-29
SC-18166 Split up and improve UJS docs

HSSalman:
We figured that it'd be good to expose the functionality and examples we offer for User Journey scripts a bit more so I did a few things in this PR:
- Split up the original UJS file into four new ones for better visibility and readability
- Modified some text to account for the new file structure
- Changed the existing links which pointed to the old file to work with the new files (in both docs and SC UI)
- Made some images look better
- Improved wording in some places

I also made [changes to the frontend](https://github.com/sematext/sematext-cloud/pull/5493) to match the ones here, as well as adding some additional fixes for content present only in the UI.


otisg (approved):
From what I can see from the diff, LGTM

PR #569, opened 2023-12-14
SC-18168 add groupby in Monitoring

cpipilas:
Add `Group By` feature in Monitoring docs. 
Link for report variables is made available in https://github.com/sematext/docs/pull/568.


docs/monitoring/reports-and-components.md
```diff
@@ -47,6 +47,13 @@ Each alert query is visible in the Alert Rules list.
 
 By checking out the [Integrations documentation](https://sematext.com/docs/integration/) you can learn more about how to monitor various programming languages, operating systems, containers (Docker, Kubernetes, Mesos), Cloud IaaS / PaaS services, iOS, Android, AWS EC2, AWS S3, Log Management, and log shippers. There you will find detailed information on how to install and configure different agents required to start sending your logs and metrics data to Sematext Cloud.
 
+### Groups
```
otisg, line 50: Groups ==> Grouping?

docs/monitoring/reports-and-components.md
```diff
@@ -47,6 +47,13 @@ Each alert query is visible in the Alert Rules list.
...
+Report components can be grouped by tags using the `Groups` drop down menu in the Monitoring Report Menu. One of the [available tags](https://sematext.com/docs/tags/common-schema/) for grouping is the `os.kernel` tag, which will put hosts running the same Linux kernel in the same group. This can be useful to check whether some hosts need to be upgraded with a newer kernel.
```
otisg, line 51: Report components ==> Data (dataseries?) in report components
drop down menu ==> drop-down? 
Monitoring Report Menu ==> monitoring report menu (because this is not a name, or is it?)

Re os.kernel.... not sure why this is brought up here?  Is it used simply as an example?

Note that this makes sense, but it's referring to data in *this particular* table.  Most tables ignore grouping, right?

But grouping is also often used for grouping data in timeseries charts.  Maybe we want to over that with text + screenshot?

otisg (changes requested):

docs/monitoring/reports-and-components.md
cpipilas: > Monitoring Report Menu ==> monitoring report menu (because this is not a name, or is it?)

We use it in this page as a name.

> Re os.kernel.... not sure why this is brought up here? Is it used simply as an example?

Yes, this is an example

> Note that this makes sense, but it's referring to data in this particular table. Most tables ignore grouping, right?

No, actually most Infra components (overview tables & charts) have grouping.

> But grouping is also often used for grouping data in timeseries charts. Maybe we want to over that with text + screenshot?

Added an example for grouping using `os.disk` tag in storage charts.

docs/monitoring/reports-and-components.md
```diff
@@ -47,10 +47,14 @@ Each alert query is visible in the Alert Rules list.
...
 
-### Groups
-Report components can be grouped by tags using the `Groups` drop down menu in the Monitoring Report Menu. One of the [available tags](https://sematext.com/docs/tags/common-schema/) for grouping is the `os.kernel` tag, which will put hosts running the same Linux kernel in the same group. This can be useful to check whether some hosts need to be upgraded with a newer kernel.
+### Grouping
+Data in report components can be grouped by tags using the `Groups` drop-down menu in the Monitoring Report Menu. For example, one of the [available tags](https://sematext.com/docs/tags/common-schema/) for grouping is the `os.kernel` tag, which will put hosts running the same Linux kernel in the same group. This can be useful to check whether some hosts need to be upgraded with a newer kernel.
```
otisg, line 51: > This can be useful to check whether some hosts need to be upgraded with a newer kernel.

In the Inventory reports, yes.  But here (looking at the screenshot) this wouldn't apply.

docs/monitoring/reports-and-components.md
```diff
@@ -47,10 +47,14 @@ Each alert query is visible in the Alert Rules list.
...
 
-![Sematext Cloud Monitoring Guide - Group By](../images/monitoring/group-by-infra.png) 
+![Sematext Cloud Monitoring Guide - Group By](../images/monitoring/group-by-infra.png)
+
+Another example is the grouping of the [infrastructure storage charts](https://apps.sematext.com/ui/infrastructure/serversStorage) by the `os.disk` tag, allowing us to quickly check for I/O spikes in specific disks.
```
otisg, line 55: Grouping by custom tags would be even better.  If web servers are tagged with `web` and db servers with `db` and then group by that tag you can see how any metric in SC is different for these two groups of servers.

otisg (approved):

docs/monitoring/reports-and-components.md
cpipilas: > In the Inventory reports, yes. But here (looking at the screenshot) this wouldn't apply.

Why not? It's convenient to group by kernel version in the Hosts Overview report and check for hosts running outdated or custom kernels, etc.

otisg: Because what I think we are showing on the screenshot I saw here yesterday is a timeseries chart where we showed some performance metric grouped by kernel.  Yes, this will let you see that you have 1 or more different kernels, but the primary goal of timeseries charts with metrics is to see performance.  If the primary interest is knowing which kernels are being used in your infra than the ideal place to go is Infra > Inventory > Servers.  That was my reasoning.

PR #598, opened 2024-02-26
SC-16400 add default ARs in integrations

cpipilas:
- This PR adds a section for default alert rules in all integrations supporting them.
- Each alert rule contains a description, an example and a list of actions to be taken. 
- Each action is limited to a short description on what the user should do.

### DIFF
--- a/docs/integration/akka.md
+++ b/docs/integration/akka.md
@@ -5,6 +5,45 @@ description: Sematext APM monitoring integration for Akka actors, dispatchers, a
 
 - Instructions: [https://apps.sematext.com/ui/howto/Akka/overview](https://apps.sematext.com/ui/howto/Akka/overview)
 
+## Akka Default Alerts
+
+As soon as you create an Akka App, you will receive a set of default [alert rules](https://sematext.com/docs/guide/alerts-guide/). These pre-configured rules will notify you of important events that may require your attention, as shown below.
+
+### CPU idle < 5%
+
+This alert rule continuously monitors the CPU idle percentage in an Akka system, triggering a warning (WARN priority) when the CPU idle falls below 5%. The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+
+Suppose an Akka system typically operates with CPU idle percentages above 5%. However, due to increased workload, the CPU idle percentage drops below 5%. When this happens, the alert rule  the alert rule triggers a warning.
+
+#### Actions to take
+
+- Review the resource usage of the Akka system to identify the cause of increased CPU usage and decreased idle percentage. This may involve analyzing application load, system configuration, and resource allocation
+
+### Network received anomaly
+
+This alert rule continuously monitors the rate of network data received by an Akka application, checking for anomalies in the received data rate over time. When anomalies are detected, it triggers a warning (WARN priority). The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+
+Suppose an Akka application typically receives network data at a consistent rate, but due to network congestion or other issues, the rate of data reception suddenly drops or spikes unexpectedly. When this happens, the alert rule triggers a warning.
+
+#### Actions to take
+
+- Check the status of the network infrastructure to determine if there are any issues such as congestion, packet loss, or connectivity problems affecting data reception
+- Review application logs and system metrics to identify any errors or warnings that may be impacting network data reception
+
+### Network transmitted anomaly
+
+This alert rule continuously monitors the rate of network transmission on the operating system level for Akka applications, checking for anomalies in transmission rates over time. When anomalies are detected, it triggers a warning (WARN priority).
+
+Suppose an Akka application typically maintains a stable rate of network transmission, but due to increased network congestion or connectivity issues, the transmission rate suddenly changes unexpectedly. When this happens, the alert rule checks for anomalies in the network transmission rate over the last hour. Upon detecting the anomaly, the alert rule triggers a warning.
+
+#### Actions to take
+
+- Check for network connectivity issues that may be affecting the transmission rate
+- Review network configurations and adjust settings to optimize network performance
+
+You can [create additional alerts](https://sematext.com/docs/alerts) on any metric.
+
+
 ## Metrics
 
 Metric Name | Key | Agg | Type | Description
--- a/docs/integration/akka.md
+++ b/docs/integration/akka.md
@@ -7,7 +7,7 @@ description: Sematext APM monitoring integration for Akka actors, dispatchers, a
 
 ## Akka Default Alerts
 
-As soon as you create an Akka App, you will receive a set of default [alert rules](https://sematext.com/docs/guide/alerts-guide/). These pre-configured rules will notify you of important events that may require your attention, as shown below.
+As soon as you create an Akka App, you will receive a set of default [alert rules](https://sematext.com/docs/guide/alerts-guide/). These pre-configured rules will [notify](https://sematext.com/docs/alerts/alert-notifications/) you of important events that may require your attention, as shown below.
 
 ### CPU idle < 5%
 
@@ -28,7 +28,7 @@ Suppose an Akka application typically receives network data at a consistent rate
 #### Actions to take
 
 - Check the status of the network infrastructure to determine if there are any issues such as congestion, packet loss, or connectivity problems affecting data reception
-- Review application logs and system metrics to identify any errors or warnings that may be impacting network data reception
+- Review application logs and system metrics for any errors or warnings that may be impacting network data reception
 
 ### Network transmitted anomaly
 
--- a/docs/integration/elasticsearch-integration.md
+++ b/docs/integration/elasticsearch-integration.md
@@ -54,7 +54,7 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
 
 ## Elasticsearch Default Alerts
 
-As soon as you create an Elasticsearch App, you will receive a set of default [alert rules](https://sematext.com/docs/guide/alerts-guide/). These pre-configured rules will notify you of important events that may require your attention, as shown below.
+As soon as you create an Elasticsearch App, you will receive a set of default [alert rules](https://sematext.com/docs/guide/alerts-guide/). These pre-configured rules will [notify](https://sematext.com/docs/alerts/alert-notifications/) you of important events that may require your attention, as shown below.
 
 ### Node count anomaly
 
@@ -104,12 +104,12 @@ Suppose an Elasticsearch cluster experiences a sudden increase in query load or 
 
 #### Actions to take
 
-- Analyze resource usage metrics for the Elasticsearch cluster, including CPU, memory, and disk utilization, to identify the source of the increased load
+- Analyze resource usage metrics for the Elasticsearch cluster, including CPU, memory, and disk utilization, to find the source of the increased load
 - Review and optimize search queries or indexing operations that may be contributing to the increased load on the cluster. Consider optimizing query performance, reducing indexing throughput
 
 ### Unassigned shards anomaly
 
-This alert rule continuously monitors the presence of unassigned shards in an Elasticsearch cluster, identifying anomalies in the number of unassigned shards over time. When anomalies are detected, it triggers a warning (WARN priority). The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+This alert rule continuously monitors the presence of unassigned shards in an Elasticsearch cluster, detecting anomalies in the number of unassigned shards over time. When anomalies are detected, it triggers a warning (WARN priority). The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
 
 Suppose an Elasticsearch cluster typically maintains a low number of unassigned shards, but due to issues such as node failures or disk space constraints, the number of unassigned shards suddenly increases. When this happens, the alert rule checks for anomalies in the number of unassigned shards over the last 30 minutes. Upon detecting the anomaly, the alert rule triggers a warning.
 
@@ -122,7 +122,7 @@ Suppose an Elasticsearch cluster typically maintains a low number of unassigned 
 
 ### Thread pool rejections anomaly
 
-This alert rule continuously monitors thread pool rejections in an Elasticsearch environment, identifying anomalies in the rate at which thread pool requests are rejected. When anomalies are detected, it triggers a warning (WARN priority). The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+This alert rule continuously monitors thread pool rejections in an Elasticsearch environment, detecting anomalies in the rate at which thread pool requests are rejected. When anomalies are detected, it triggers a warning (WARN priority). The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
 
 Suppose an Elasticsearch cluster experiences a sudden increase in thread pool rejections, potentially due to resource limitations or high query loads. When this happens, the alert rule checks for anomalies in thread pool rejections over the last 90 minutes. Upon detecting the anomaly, the alert rule triggers a warning.
 
@@ -153,7 +153,7 @@ Suppose there is some activity detected in the swap usage on a node in the Elast
 
 #### Action to take
 
-- Review system resource metrics to identify any spikes or anomalies in CPU, memory, or disk usage that may be contributing to swap usage
+- Review system resource metrics to find any spikes or anomalies in CPU, memory, or disk usage that may be contributing to swap usage
 - Evaluate Elasticsearch configuration settings and adjust resource allocations as needed to optimize performance and prevent any swap usage
 
 ### Open files > 85%
@@ -164,7 +164,7 @@ Suppose an Elasticsearch cluster typically operates with a healthy percentage of
 
 #### Actions to take
 
-- Review resource usage metrics and system logs to identify any issues contributing to the high percentage of open files
+- Review resource usage metrics and system logs to find any issues contributing to the high percentage of open files
 - Review Elasticsearch cluster configuration settings and consider optimizing resource allocation and file management settings to better handle file usage and prevent excessive file opening
 
 ### Load average
--- a/docs/integration/elasticsearch-integration.md
+++ b/docs/integration/elasticsearch-integration.md
@@ -52,6 +52,134 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
 
 ![Dropping nodes and relocation of shards](https://sematext.com/wp-content/uploads/2019/03/Screen-Shot-2019-03-26-at-10.59.52-1.png)
 
+## Elasticsearch Default Alerts
+
+As soon as you create an Elasticsearch App, you will receive a set of default [alert rules](https://sematext.com/docs/guide/alerts-guide/). These pre-configured rules will notify you of important events that may require your attention, as shown below.
+
+### Node count anomaly
+
+This alert rule continuously monitors the count of nodes in an Elasticsearch cluster, checking for anomalies in the number of nodes present within the cluster. When anomalies are detected, it triggers a warning (WARN priority). The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+
+Suppose an Elasticsearch cluster typically maintains a stable number of nodes, but due to various factors such as node failures, scaling activities, or network issues, the node count experiences sudden changes. When this happens, the alert rule checks for anomalies in the count of nodes over the last 90 minutes. Upon detecting the anomaly, the alert rule triggers a warning.
+
+#### Actions to take
+
+- Check the status of nodes within the Elasticsearch cluster for any nodes that may be offline, unavailable, or experiencing issues
+- If node failures are detected, you may need to restart failed nodes or replace hardware
+- If the node count changes due to scaling activities (e.g., adding or removing nodes), review the recent scaling events to confirm that they are intentional and expected
+- Monitor network connectivity between nodes within the Elasticsearch cluster for any network issues that may be affecting communication and node discovery
+
+
+### Java old gen usage > 97%
+
+This alert rule continuously monitors the usage of Java's old generation heap memory in an Elasticsearch environment, triggering a warning if the usage exceeds 97%. The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+
+Suppose the Java old generation heap memory usage in the Elasticsearch environment starts increasing and eventually exceeds 97% over a 5-minute period.
+
+#### Actions to take
+
+- Analyze the application's memory usage patterns for any memory leaks
+- Review and optimize the Java Virtual Machine (JVM) configuration, including heap size settings, garbage collection algorithms, and memory management parameters
+- Monitor system resources, including memory, CPU, and disk I/O, and consider scaling them up if necessary
+- Investigate recent application changes, updates, or deployments that may have contributed to the spike in memory usage
+
+### Field data size
+
+This alert rule continuously monitors the field data size in an Elasticsearch cluster and triggers a warning if the field data size exceeds a certain threshold (20 in this case). The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+
+Suppose the field data size in the Elasticsearch cluster increases significantly due to increased query loads or inefficient queries. When this happens, the alert rule checks the field data size over the last 10 minutes. If it exceeds 20% of the JVM heap committed memory during this timeframe, the alert rule triggers a warning.
+
+#### Actions to take
+
+- Analyze the query patterns and usage that contribute to the increased field data size. Look for inefficient or resource-intensive queries that may need optimization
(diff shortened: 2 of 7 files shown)


docs/integration/opensearch-integration.md
```diff
@@ -40,6 +40,135 @@ OpenSearch runs within a Java Virtual Machine (JVM) and [monitoring JVM](https:/
...
+#### Actions to take
+
+- Investigate the cause of increased memory usage in the OpenSearch cluster, such as processes or activities contributing to the high memory consumption
+- Review and optimize the configuration settings of the OpenSearch cluster, including heap size allocation and cache settings
+
+### Swap usage
+
+This alert rule continuously monitors swap usage in an OpenSearch environment by tracking the rate of swap input/output operations. When swap usage exceeds the specified threshold, it triggers a warning (WARN priority). The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+
+Suppose the swap usage on a node in the OpenSearch cluster suddenly spikes, indicating potential resource constraints or performance issues. When this happens, the alert rule checks for swap usage over the last 10 minutes. The alert is triggered when the sum of swap input/output operations within the specified timeframe is more than 1 (the threshold value).
```
cpipilas, line 139: @radu-gheorghe I think I'm wrong here. Can you confirm the condition for triggering this alert?

radu-gheorghe: Actually, I think any swap usage (even if you read 1 byte) would trigger the alert. The idea is that OpenSearch needs OS cache (read: free memory) to work with the index. If the OS swaps the JVM memory, it means one of the following: there's not enough RAM on the box or (more commonly) the OS decides to swap a chunk of RAM preemptively (because of an inappropriate `vm.swappiness` setting. Either way, it could introduce significant (read: unacceptable) delays when OpenSearch wants to access that piece of memory, so it's a no-no to swap.

cpipilas: Thanks @radu-gheorghe! I've updated the AR.

docs/integration/akka.md
```diff
@@ -5,6 +5,45 @@ description: Sematext APM monitoring integration for Akka actors, dispatchers, a
...
+This alert rule continuously monitors the rate of network data received by an Akka application, checking for anomalies in the received data rate over time. When anomalies are detected, it triggers a warning (WARN priority). The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+
+Suppose an Akka application typically receives network data at a consistent rate, but due to network congestion or other issues, the rate of data reception suddenly drops or spikes unexpectedly. When this happens, the alert rule triggers a warning.
+
+#### Actions to take
+
+- Check the status of the network infrastructure to determine if there are any issues such as congestion, packet loss, or connectivity problems affecting data reception
+- Review application logs and system metrics for any errors or warnings that may be impacting network data reception
+
+### Network transmitted anomaly
```
radu-gheorghe, line 33: This is fine, but maybe we can just say it's the same as ingress, but this is egress? :)

docs/integration/elasticsearch-integration.md
```diff
@@ -52,6 +52,136 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
...
+
+### Node count anomaly
+
+This alert rule continuously monitors the count of nodes in an Elasticsearch cluster, checking for anomalies in the number of nodes present within the cluster. When anomalies are detected, it triggers a warning (WARN priority). The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+
+Suppose an Elasticsearch cluster typically maintains a stable number of nodes, but due to various factors such as node failures, scaling activities, or network issues, the node count experiences sudden changes. When this happens, the alert rule checks for anomalies in the count of nodes over the last 90 minutes. Upon detecting the anomaly, the alert rule triggers a warning.
+
+#### Actions to take
+
+- Check the status of nodes within the Elasticsearch cluster for any nodes that may be offline, unavailable, or experiencing issues
```
radu-gheorghe, line 67: You'll probably want to check the logs of the node that went down and/or logs of the master node.

docs/integration/elasticsearch-integration.md
```diff
@@ -52,6 +52,136 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
...
+- If node failures are detected, you may need to restart failed nodes or replace hardware
+- If the node count changes due to scaling activities (e.g., adding or removing nodes), review the recent scaling events to confirm that they are intentional and expected
+- Monitor network connectivity between nodes within the Elasticsearch cluster for any network issues that may be affecting communication and node discovery
+
+
+### Java old gen usage > 97%
+
+This alert rule continuously monitors the usage of Java's old generation heap memory in an Elasticsearch environment, triggering a warning if the usage exceeds 97%. The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+
+Suppose the Java old generation heap memory usage in the Elasticsearch environment starts increasing and eventually exceeds 97% over a 5-minute period.
```
radu-gheorghe, line 77: I would add that this shouldn't happen in a healthy environment. It's likely that the node will either face an out of memory exception or run into the Parent Circuit Breaker. Either way, you'd have unexpected failures.

docs/integration/elasticsearch-integration.md
```diff
@@ -52,6 +52,136 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
...
+
+#### Actions to take
+
+- Analyze the application's memory usage patterns for any memory leaks
+- Review and optimize the Java Virtual Machine (JVM) configuration, including heap size settings, garbage collection algorithms, and memory management parameters
```
radu-gheorghe, line 82: Link to some of our blog posts on the topic?

docs/integration/elasticsearch-integration.md
```diff
@@ -52,6 +52,136 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
...
+- Monitor system resources, including memory, CPU, and disk I/O, and consider scaling them up if necessary
```
radu-gheorghe, line 83: I'd remove this, we're already monitoring :)

docs/integration/elasticsearch-integration.md
```diff
@@ -52,6 +52,136 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
...
+- Investigate recent application changes, updates, or deployments that may have contributed to the spike in memory usage
+
+### Field data size
+
+This alert rule continuously monitors the field data size in an Elasticsearch cluster and triggers a warning if the field data size exceeds a certain threshold (20 in this case). The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+
+Suppose the field data size in the Elasticsearch cluster increases significantly due to increased query loads or inefficient queries. When this happens, the alert rule checks the field data size over the last 10 minutes. If it exceeds 20% of the JVM heap committed memory during this timeframe, the alert rule triggers a warning.
```
radu-gheorghe, line 90: This is getting repetitive. In this particular case, I'd say that significant field data usage points to a misconfiguration. Normally, you'd only use field data for global ordinals. If you're using >20%, you probably do sorting/aggregations on a text field with field data enabled. Which is dangerous (you may run out of heap on an expensive query). So you'd want to pre-process the data in the pipeline before ES and do your sorting/aggregations on doc_values instead.

docs/integration/akka.md
cpipilas: Yep, I'm trying to give all the details for each alert rule. There is a bunch of repeated text, but it's like each piece of info stands alone and helps connect with SC later on.

docs/integration/elasticsearch-integration.md
```diff
@@ -52,6 +52,136 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
...
+
+#### Actions to take
+
+- Analyze the query patterns and usage that contribute to the increased field data size. Look for inefficient or resource-intensive queries that may need optimization
```
radu-gheorghe, line 94: _cat/fielddata will tell you which fields use more field data.

docs/integration/elasticsearch-integration.md
```diff
@@ -52,6 +52,136 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
...
+- Optimize queries to reduce the amount of field data loaded into memory. This may involve restructuring queries, using filters, or optimizing aggregations to minimize memory usage
```
radu-gheorghe, line 95: You can't fix this by changing queries, see my earlier comment.

docs/integration/elasticsearch-integration.md
```diff
@@ -52,6 +52,136 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
...
+- Consider scaling up the resources allocated to the Elasticsearch cluster, such as increasing the JVM heap size, to accommodate the increased field data size
+
+
+### Tripped parent circuit breaker
```
radu-gheorghe, line 99: Parent Breaker has two modes:
- the default, on "real data". Meaning it looks at the current heap usage before doing an operation. If heap usage > threshold --> operation is not run, because you're in danger of running out of heap
- real_data=false --> just the sum of all the other breakers. I.e. if an operation tries to load memory that would use more than X% of heap across all breakers (e.g. field data, in-flight requests...), that operation isn't run

docs/integration/elasticsearch-integration.md
```diff
@@ -52,6 +52,136 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
...
+
+This alert rule continuously monitors the tripping of the parent circuit breaker in an Elasticsearch cluster, detecting instances where the circuit breaker has been triggered due to resource constraints or overload. When such instances are detected, it triggers a warning (WARN priority). The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+
+Suppose an Elasticsearch cluster experiences a sudden increase in query load or indexing throughput, leading to resource contention and triggering the parent circuit breaker. When this happens, the alert rule checks for instances of the parent circuit breaker being tripped over the last 5 minutes. The alert is triggered as soon as the circuit breaker is tripped at least once within the specified timeframe.
+
+#### Actions to take
+
```
radu-gheorghe, line 106: More heap, nodes, too?

docs/integration/elasticsearch-integration.md
```diff
@@ -52,6 +52,136 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
...
+- Analyze resource usage metrics for the Elasticsearch cluster, including CPU, memory, and disk utilization, to find the source of the increased load
+- Review and optimize search queries or indexing operations that may be contributing to the increased load on the cluster. Consider optimizing query performance, reducing indexing throughput
+
+### Unassigned shards anomaly
+
+This alert rule continuously monitors the presence of unassigned shards in an Elasticsearch cluster, detecting anomalies in the number of unassigned shards over time. When anomalies are detected, it triggers a warning (WARN priority). The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+
+Suppose an Elasticsearch cluster typically maintains a low number of unassigned shards, but due to issues such as node failures or disk space constraints, the number of unassigned shards suddenly increases. When this happens, the alert rule checks for anomalies in the number of unassigned shards over the last 30 minutes. Upon detecting the anomaly, the alert rule triggers a warning.
```
radu-gheorghe, line 114: Normally, you shouldn't have unassigned shards. But OK, if you have some, at least the number shouldn't change :)

docs/integration/elasticsearch-integration.md
```diff
@@ -52,6 +52,136 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
...
+- Check the status of Elasticsearch nodes to determine if any nodes are experiencing issues or are offline
+- Review disk space on Elasticsearch nodes to see if there is sufficient space available for shard allocation
+- Review shard allocation settings in the Elasticsearch cluster configuration to make sure that shards are allocated properly and evenly across nodes
+- Recover unassigned shards and allocate them to available nodes in the cluster
+
+### Thread pool rejections anomaly
+
+This alert rule continuously monitors thread pool rejections in an Elasticsearch environment, detecting anomalies in the rate at which thread pool requests are rejected. When anomalies are detected, it triggers a warning (WARN priority). The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+
+Suppose an Elasticsearch cluster experiences a sudden increase in thread pool rejections, potentially due to resource limitations or high query loads. When this happens, the alert rule checks for anomalies in thread pool rejections over the last 90 minutes. Upon detecting the anomaly, the alert rule triggers a warning.
```
radu-gheorghe, line 127: A healthy cluster should have none, but maybe this is normal e.g. you're ingesting at the limit of what the cluster can handle. In this case, we should have at least a relatively constant number

docs/integration/elasticsearch-integration.md
```diff
@@ -52,6 +52,136 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
...
+
+#### Actions to take
+
+- Review system metrics for the Elasticsearch cluster, including CPU, memory, and disk usage, for any resource constraints that may be contributing to thread pool rejections
+- Analyze query patterns for any inefficient or resource-intensive queries. Optimize queries to reduce the load on the cluster
```
radu-gheorghe, line 132: This only applies for the search thread pool

docs/integration/elasticsearch-integration.md
```diff
@@ -52,6 +52,136 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
...
+- Review thread pool configurations and adjust settings such as thread counts, queue sizes, and timeouts to better accommodate the workload and prevent thread pool saturation
```
radu-gheorghe, line 133: This is tricky advice, because we don't want to tell people to increase queues and just postpone the problem. But maybe we can tell people to check the calling apps and use fewer threads to talk to Elasticsearch. On the indexing side, exponential backoff is a good practice.

docs/integration/elasticsearch-integration.md
```diff
@@ -52,6 +52,136 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
...
+
+### Used memory > 80%
+
+This alert rule continuously monitors memory usage in an Elasticsearch environment and triggers a warning (WARN priority) when the used memory exceeds 80% of the total available memory. The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+
+Suppose the Elasticsearch cluster experiences a sudden increase in activity or data volume, leading to increased memory usage. When the used memory exceeds 80% of the total available memory, the alert rule checks memory usage over the last hour. Upon crossing the threshold, the alert rule triggers a warning.
```
radu-gheorghe, line 139: Data volume shouldn't increase memory usage. Elasticsearch should use heap + a few other things. If there's more than 80% memory used => either heap is set too high or something else is using memory. Or some weird bug.

docs/integration/elasticsearch-integration.md
```diff
@@ -52,6 +52,136 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
...
+
+#### Actions to take
+
+- Investigate the cause of increased memory usage in the Elasticsearch cluster, such as processes or activities contributing to the high memory consumption
+- Review and optimize the configuration settings of the Elasticsearch cluster, including heap size allocation and cache settings
```
radu-gheorghe, line 144: Just heap size. Cache doesn't matter.

docs/integration/elasticsearch-integration.md
```diff
@@ -52,6 +52,136 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
...
+
+### Swap usage
+
+This alert rule continuously monitors swap usage in an Elasticsearch environment by tracking the rate of swap input/output operations. When any amount of swap usage is detected, it triggers a warning (WARN priority). This includes even the slightest swap activity, such as reading or writing a single byte to or from swap space.
+
+The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+
+Suppose there is some activity detected in the swap usage on a node in the Elasticsearch cluster. Despite the relatively small amount of swap activity, the alert rule triggers a warning to prevent any big (and potentially unacceptable) slowdowns in Elasticsearch caused by accessing swap memory.
```
radu-gheorghe, line 152: Yeah, you don't want to swap heap memory -> things can get very slow.

docs/integration/elasticsearch-integration.md
```diff
@@ -52,6 +52,136 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
...
+
+#### Action to take
```
radu-gheorghe, line 154: Just disable swap, that's the only thing that can be done. Well, flavors: swapoff or vm.swappiness=1

docs/integration/elasticsearch-integration.md
```diff
@@ -52,6 +52,136 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
...
+- Review system resource metrics to find any spikes or anomalies in CPU, memory, or disk usage that may be contributing to swap usage
+- Evaluate Elasticsearch configuration settings and adjust resource allocations as needed to optimize performance and prevent any swap usage
+
+### Open files > 85%
+
+This alert rule continuously monitors the percentage of open files in an Elasticsearch cluster. When the percentage exceeds 85% within the specified timeframe, it triggers a warning (WARN priority). The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+
+Suppose an Elasticsearch cluster typically operates with a healthy percentage of open files, but due to increased usage or resource limitations, the percentage of open files exceeds 85%. When this happens, the alert rule checks for instances where the percentage of open files exceeds 85% within the last 10 minutes and triggers a warning.
+
+#### Actions to take
```
radu-gheorghe, line 165: Make sure the limit of open files is correct (65K, I think, is recommended). If you get anywhere close, you have a big problem: weird merge policy, tons of very small shards maybe... never saw that situation. Changing directory (e.g. to hybridfs) should help, as well as increasing compound format percentage, force-merging old data.... but again, 99% it's a misconiguration.

docs/integration/elasticsearch-integration.md
```diff
@@ -52,6 +52,136 @@ Last, but certainly not least, you may want to get an alert if a node leaves the
...
+
+### Load average
+
+This alert rule continuously monitors the load average of an Elasticsearch cluster and triggers a warning when the load average exceeds a specified threshold (currently when load average is more than 2). The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+
+Suppose the average load on the Elasticsearch cluster typically remains below 2, but due to increased query loads or resource constraints, the load average spikes above 2. When this happens, the alert rule checks for load average values over the last 30 minutes. Upon detecting the load average anomaly, the alert rule triggers a warning.
+
+#### Actions to take
+
+-  Review system metrics such as CPU, memory, and disk usage for any source of increased load on the Elasticsearch cluster
```
radu-gheorghe, line 178: not memory or disk, but everything related to load: thread pools, indexing and search operations, heap usage...

docs/integration/elasticsearch-integration.md
cpipilas: Thanks Radu. So we alert on any Parent circuit breaker exception. If real memory is used it will be triggered at 95% of JVM heap, otherwise at 70% due to estimates. I guess both are configurable (?). Do you think we need to clarify anything here? Just mention the settings maybe?

docs/integration/kafka.md
```diff
@@ -161,6 +161,104 @@ The Sematext Kafka monitoring agent collects the following metrics.
...
+### Consumer lag anomaly
+
+This alert rule continuously monitors consumer lag in Kafka consumer clients, detecting anomalies in lag values over time. When anomalies are detected, it triggers a warning (WARN priority). The minimum delay between consecutive notifications triggered by this alert rule is set to 10 minutes.
+
+Suppose a Kafka consumer client typically processes messages without significant lag, but due to increased message volume or processing issues, the consumer lag suddenly spikes beyond normal levels. When this happens, the alert rule checks for anomalies in consumer lag values over the last 90 minutes and triggers a warning.
+
+#### Actions to take
+
+- Review Kafka consumer client configurations, message processing logic and system metrics to find the underlying cause of the increased lag
+- If the increased lag is due to limited resources, consider scaling up Kafka consumer client resources such as CPU, memory, or network bandwidth
```
radu-gheorghe, line 177: Or increase the number of partitions and consumers?

docs/integration/kafka.md
```diff
@@ -161,6 +161,104 @@ The Sematext Kafka monitoring agent collects the following metrics.
...
+
+Suppose a Kafka cluster typically maintains full replication of partitions across all brokers, but due to broker failures, network issues, or limited resources, some partitions become under-replicated. When this happens, the alert rule checks for anomalies in the count of under-replicated partitions over the last 10 minutes. Upon detecting the anomaly in the count of under-replicated partitions, the alert rule triggers a warning.
+
+#### Actions to take
+
+- Check the status of Kafka brokers for any broker failures or issues that may be causing partitions to become under-replicated
+- Review the replication configurations for affected topics and partitions and double check that replication factors are set correctly
+- AAddress any network issues or connectivity problems between Kafka brokers that may be impacting the replication of partitions
```
radu-gheorghe, line 227: s/AAddress/Address/

docs/integration/kafka.md
```diff
@@ -161,6 +161,104 @@ The Sematext Kafka monitoring agent collects the following metrics.
...
+
+Suppose a Kafka cluster typically maintains all partitions online and available for data replication and consumption. However, due to broker failures, disk failures, or other issues, some partitions become offline and unavailable. When this happens, the alert rule checks for anomalies in the count of offline partitions over the last 10 minutes. Upon detecting the anomaly, the alert rule triggers a warning.
+
+#### Actions to take
+
+- Check the status of Kafka brokers for any broker failures or issues that may be causing partitions to become offline
+- Review the replication status and configuration of affected partitions to understand why they became offline. Determine if there are any replication issues or failures
+- If offline partitions are due to broker failures, try to recover or replace the failed brokers and make sure that partitions are redistributed and replicated properly
+- If offline partitions are due to disk failures or storage issues, address the underlying disk failures and mae sure that Kafka data directories are accessible and properly configured
```
radu-gheorghe, line 242: s/mae/make/

docs/integration/opensearch-integration.md
```diff
@@ -40,6 +40,137 @@ OpenSearch runs within a Java Virtual Machine (JVM) and [monitoring JVM](https:/
 
 Finally, high disk reads and writes can indicate a poorly tuned system. Since accessing the disk is an expensive process in terms of time, a well-tuned system should reduce disk I/O wherever possible.
 
+
+## OpenSearch Default Alerts
```
radu-gheorghe, line 44: I assume those are the same as Elasticsearch for the moment.

docs/integration/solr-integration.md
```diff
@@ -97,6 +97,24 @@ The Sematext Solr monitoring agent collects the following metrics.
...
+### Warmup time > 5m
+
+This alert rule continuously monitors the warmup time of a Solr cache, identifying instances where the warmup time exceeds 5 minutes. When such instances are detected, it triggers a warning (WARN priority). The minimum delay between consecutive alerts triggered by this alert rule is set to 10 minutes.
+
+Suppose a Solr cache is configured to warm up before serving requests, and the typical warmup time is around 3 minutes. However, due to increased data volume or inefficient warmup processes, the warmup time suddenly increases to 7 minutes. When this happens, the alert rule checks for warmup time anomalies over the last 5 minutes. Upon detecting the warmup time anomaly, the alert rule triggers a warning.
+
+#### Actions to take
+
+- Review the warmup process of the Solr cache for any bottlenecks or configuration issues contributing to the extended warmup time
+- Optimize the warmup process by optimizing queries or adjusting cache configuration settings to reduce warmup time
```
radu-gheorghe, line 113: You'd have to adjust autoWarmCount for some caches or the newSearcher queries. But maybe it's also that people do soft commits too often (either autoSoftCommit or from the application).

docs/integration/akka.md
radu-gheorghe: Got it. Thanks!

docs/integration/elasticsearch-integration.md
radu-gheorghe: Right, 95% is the current default for real memory, 70% for "sum-of-all-circuit-breakers". I think we can safely assume people have real memory enabled, I didn't really see clusters without real memory. And yes, it's configurable.

As for clarifications, I think it's important to say that the parent CB triggered implies that the memory usage right now is very high.

radu-gheorghe (changes requested):
I've done a full pass and left some comments, though I see you resolved most of them. Let me know if you need another pass :)

docs/integration/opensearch-integration.md
cpipilas: Yes, that's  right

cpipilas:
> I've done a full pass and left some comments, though I see you resolved most of them. Let me know if you need another pass :)

Thank you, thank you! I think I've addressed all the comments, I've just requested a review to make sure there are no more comments or clarifications.

radu-gheorghe:
LGTM, thanks for adjusting!

radu-gheorghe (approved):

PR #599, opened 2024-03-06
SC-18648 add release notes for sematext-agent 3.3.0

boratanrikulu:


docs/agents/sematext-agent/releasenotes.md
```diff
@@ -39,6 +39,19 @@ NULL
 
 -->
 
+## Version 3.3.0
+
+Date: March 5, 2024
+
+### Improvements
+
+- Updated the App-Agent command signature to avoid orphan processes after a OOM happen by the kernel.
```
yago-123, line 48: I think this is wrong because Otis corrected me during last demo: 

- `after a OOM happen by the kernel.` -> `after a OOM happen by the JVM.`

Slightly better: `Updated the App-Agent command signature to prevent orphan processes after an Out-of-Memory (OOM) event occurs in the JVM.`

cpipilas (approved):
We are good now

PR #604, opened 2024-04-04
Ful xsell url updates

fulyauluturk:
Changed URLs `troubleshoot` to `root cause discovery`


fulyauluturk:
Created https://github.com/sematext/sematext-cloud/pull/5859 for SC updates. CC @otisg

PR #605, opened 2024-04-09
SC-19179 reconfigure agent to send metrics to another infra app

cpipilas:
Add a section in Monitoring FAQ to explain how to reconfigure Sematext Agent to send metrics to another Infra App


cpipilas:
> LGTM. Ofc, in large deployments there won't be "connect to the host"...

Hmm right, I will remove it. If one is using Ansible, for example, this doesn't apply.

PR #631, opened 2024-07-11
SC-19827 add windows services page

cpipilas:
- Add a new page for Windows Services monitoring
    - Metrics & Tags
    - Reports (will be updated in a new PR)
    - Alerts

### DIFF
--- a/docs/monitoring/windows-services.md
+++ b/docs/monitoring/windows-services.md
@@ -1,0 +1,74 @@
+title: Sematext Windows Services Monitoring
+description: Monitoring of Windows Services provides real-time insights into their status, performance, and potential issues
+
+Sematext Monitoring now includes support for Windows Services. Starting from Sematext Agent [version 3.5.0](https://sematext.com/docs/agents/sematext-agent/releasenotes/#version-350), you can collect and monitor the status of any Windows Service available in your system.
+
+## Metrics
+
+The [Sematext Agent for Windows](https://sematext.com/docs/agents/sematext-agent/windows-installation/), starting from version 3.5.0, is capable of collecting metrics related to the status of Windows services. At present, the primary metric supported is the status of each service including a number of statistics, which allows users to determine whether a service is running, stopped, or experiencing issues.
+
+
+### List of available metrics
+
+|Name| Type  | Description |
+|----|-------|-------------|
+|windows.service.status            | number |status of a Windows service|
+|windows.service.total             | gauge  |total number of Windows services|
+|windows.service.total.running     | gauge  |total number of running Windows services|
+|windows.service.total.not_running | gauge  |total number of non running Windows services|
+
+
+### Available Windows Services statuses 
+
+Below is a list of all the possible values (statuses) of the `windows.service.status` metric.
+
+|Value | Service Status  |
+|------|-----------------|
+|1     | stopped         |
+|2     | start pending   |
+|3     | stop pending    |
+|4     | running         |
+|5     | continue pending|
+|6     | pause pending   |
+|7     | paused          |
+
+### Available Tags
+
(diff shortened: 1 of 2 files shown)


docs/monitoring/windows-services.md
```diff
@@ -0,0 +1,74 @@
...
+## Reports
+
+Coming Soon!
+
+
+## Default Alerts
+
+A set of pre-defined [alerts](https://sematext.com/docs/alerts/) is supported for the status of several important Windows services. These alerts are triggered when the status of any monitored service is not running. The following services are monitored out-of-the-box:
+
+- Windows Time (W32Time)
```
fulyauluturk, line 57: @cpipilas you think we should add a short sentence for each explaining what it does and why it is important?

fulyauluturk:
Added a minor comment otherwise LGTM

docs/monitoring/windows-services.md
cpipilas: Aha, all these alerts do exactly the same thing, Fulya. They are checking each service status. When the status is anything but `running`, they are triggered. :)

fulyauluturk: Yes I meant what each service does, to explain why did we create default alert for that service specifically why it is important.

docs/monitoring/windows-services.md
```diff
@@ -0,0 +1,74 @@
...
+|2     | start pending   |
+|3     | stop pending    |
+|4     | running         |
+|5     | continue pending|
+|6     | pause pending   |
+|7     | paused          |
+
+### Available Tags
+
+Below is a list of available [tags](https://sematext.com/docs/tags/) for the `windows.service.status` metric. Tags can be used to retrieve individual service names or process IDs related with a specific service. Edit any chart in the Infrastructure > Windows > Top Services report to understand how you can use a the `windows.service.status`, grouped by a tag.
```
otisg, line 37: related with --> related to

docs/monitoring/windows-services.md
```diff
@@ -0,0 +1,74 @@
...
+
+|Tag name | Description  |
+|---------|--------------|
+|windows.service.name   | name of the Windows service       |
+|windows.service.type   | type of the Windows service       |
+|windows.service.pid    | process ID of the Windows service |
+
+For more information about the status of Windows Services and the available service types, please refer to the [official documentation](https://learn.microsoft.com/en-us/windows/win32/api/winsvc/ns-winsvc-service_status).
```
otisg, line 45: Why does "available service types" matter here if STAW monitors only specific services?  If one has service Foo and Foo is not monitored by STAW by default, can one do something to tell Foo to start monitoring Foo?

otisg (approved):

docs/monitoring/windows-services.md
boratanrikulu: > if STAW monitors only specific services?

STAW monitors all services. As I know, default charts and default alerts have specific services but users can create their own charts or alerts for all services.

docs/monitoring/windows-services.md
cpipilas: > Yes I meant what each service does, to explain why did we create default alert for that service specifically why it is important.

Oh, I see. This will be added when we document the upcoming Windows Services reports. There, we will have the chance to state the expected status of each service, whether it should run continuously, and what the consequences will be if the service is not running.

otisg: @cpipilas Considered doing that in form of short tips in the report? Or maybe we would end up with too many components on a single report?

cpipilas: Right, I haven't found a nice representation yet for tips (we are out of real estate) but I will check what we can do

PR #637, opened 2024-08-07
SC-19984 Rewrite Synthetics UJS examples which used Live Demos

HSSalman:
Some of our Synthetics Browser monitor UJS examples used our live demos. Since that functionality has been removed, these have been rewritten to keep the examples relevant and up-to-date. I have also tested that these new examples worked fine for multiple runs from all our predefined locations, and added the missing description for the _Multiple Page Load Times_ example to our docs.


hasan-wajahat (approved):
Good work 👏 

In the past Airbnb has caused me some pains, they change their selectors and UI a lot and it was difficult to handle some of their inputs. But its working for your case so should be fine.

PR #663, opened 2025-01-17
SC-20345 add release notes for sematext-agent v3.8.0

boratanrikulu:


fulyauluturk (approved):

fulyauluturk:
@boratanrikulu LGTM but note that we cannot deploy docs right now. Otis is looking into it.

PR #676, opened 2025-05-06
SC-19802 Add CI/CD Monitors, CI/CD Groups and Dynamic URLs docs

HSSalman:
Docs covering the upcoming Synthetics GitHub CI/CD integration. I was going to include the installation instructions too since I have notes on them, but while writing them out I didn't really like the added complexity. I'll try some ideas tomorrow to see if those can simplify things - it's mainly the GH API that's introducing the complexity, and there may be a way around it through an official Sematext Synthetics CI/CD App.

Meanwhile, @fulyauluturk can you please take a look at the current content to see if anything needs clearing up? I think I covered pretty much everything on the SC side, but I'd appreciate some extra input.


fulyauluturk:
@HSSalman It is a little bit unclear that once you set that `<DYNAMIC_URL>` how you should set things up for each run to replace the placeholder automatically. For example, how will each run know to replace `<DYNAMIC_URL>/health` with `pr-1227.test.myapp.com/health` and how it'll increase the PR number automatically for each run. This should be configured within the GH action script, right? Shouldn't we state that and maybe link it to installation instructions once we have that page.

docs/synthetics/ci-cd/ci-cd-installation.md
```diff
@@ -0,0 +1,283 @@
...
+
+permissions:
+  contents: read   # Allows us to read the URL from the deployment event
+
+jobs:
+  sematext_synthetics_check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get Vercel Deployment URL
+        id: vercel_deployment_url
```
otisg, line 88: @HSSalman if I don't use Vercel, what do I put here? (what a user might be wondering)

docs/synthetics/ci-cd/ci-cd-installation.md
```diff
@@ -0,0 +1,283 @@
...
+        run: |
+          echo "DEPLOYMENT_URL=$(cat "$GITHUB_EVENT_PATH" | jq .deployment_status.target_url)" >> $GITHUB_ENV
+
+      - name: Run Sematext Synthetics CI/CD Integration
+        id: sematext_check
+        uses: hssalman/cicd-repo-action@v0.9.3
```
otisg, line 94: @HSSalman is this the actual thing one will have to use or will this become sematext/.... ?

docs/synthetics/ci-cd/ci-cd-installation.md
```diff
@@ -0,0 +1,283 @@
...
+
+Add the following to your GitHub workflow file:
+
+```yaml
+steps:
+  - name: Run Sematext Synthetics Check
+    uses: sematext/synthetics-cicd-action@v1
+    with:
+      MONITOR_GROUP_ID: 42
+      REGION: "US"
```
otisg, line 32: @HSSalman nitpick - single quotes to match the rest here, and how I see region specified in the YML example below

docs/synthetics/ci-cd/ci-cd-installation.md
```diff
@@ -0,0 +1,283 @@
...
+          # Process the result to ensure proper formatting in GitHub Markdown
+          if [[ -n "${{ steps.sematext_check.outputs.result }}" ]]; then
+            FORMATTED_RESULT=$(echo '${{ steps.sematext_check.outputs.result }}' | sed 's/\\n/\n/g')
+            RESULTS_BLOCK="<pre>${FORMATTED_RESULT}</pre>"
+          else
+            RESULTS_BLOCK=""
+          fi
+
+          if [[ "${{ steps.sematext_check.outputs.status }}" == "passed" ]]; then
+            gh api "repos/${{ env.REPO }}/check-runs/${{ env.check_run_id }}" -X PATCH \
```
otisg, line 232: @HSSalman this "gh api" is OK for everyone to use?

docs/synthetics/ci-cd/ci-cd-installation.md
```diff
@@ -0,0 +1,283 @@
...
+              -F "name=Sematext CI/CD Test" \
+              -F "output[title]=Sematext Synthetics Check Passed" \
+              -F "output[summary]=All monitors finished successfully." \
+              -F "output[text]=$JOB_DETAILS<br>${RESULTS_BLOCK}" \
+              -F "status=completed" \
+              -F "conclusion=success"
+          else
+            FAILURE_REASON="Some monitors have failed."
```
otisg, line 240: @HSSalman we use "monitors" and "checks" interchangeably. Won't be confusing?

docs/synthetics/ci-cd/ci-cd-monitors.md
```diff
@@ -0,0 +1,84 @@
...
+
+You'll notice that you'll no longer be able to select the interval between automatically scheduled runs for a **CI/CD Monitor**. This is because these monitors run only when you explicitly send a run request, meaning you'll execute them on-demand and avoid cluttering their run results with potentially irrelevant automatically scheduled runs. The idea here is to use Synthetics monitors as regression tests for your projects by integrating them into your CI/CD pipeline, thus catching issues before they make it to production. To reflect this lack of a regularly scheduled interval, these monitors are marked with a special badge in the **Interval** column of the Monitors Overview table.
+
+![CI/CD Monitors in the Monitors Overview Table](/docs/images/synthetics/cicd-monitors-overview.png)
+
+
+
+## CI/CD Groups
+
+When running tests for your projects, you'll most likely want to test a number of endpoints and User Journeys. To facilitate running all the **CI/CD Monitors** you need at once, you can group them together in **CI/CD Groups**. These can be created and viewed in the Synthetics sidebar.
```
otisg, line 28: @HSSalman "CI/CD Groups" or "CI/CD Monitor Groups" or just "Monitor Groups"? Aha, looks like we use "CI/CD Groups" a lot elsewhere. OK.

docs/synthetics/ci-cd/ci-cd-installation.md
HSSalman: It'll be an official `sematext` repo, yes. This is just a temporary one until we're satisfied with how things work and I left it in so that Fulya can easily try out these install guides.

docs/synthetics/ci-cd/ci-cd-installation.md
HSSalman: Yes, it's the official [GitHub CLI that's used in workflows](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-github-cli-in-workflows)

docs/synthetics/ci-cd/ci-cd-installation.md
HSSalman: Hmm... right. We sort of have to use both because we're talking about GitHub Check Runs and our own Synthetics Monitors. I'll take another look to see if I can make things more consistent, but we'll still have both terms.

HSSalman:
These should be the final versions unless there's some feedback that says otherwise.

However, @otisg @fulyauluturk we still need to decide what we'll do with the old pages, since they point to some workflows which can only run one monitor per action, which doesn't seem useful and will only confuse users with multiple variants of what they might think is the same action.

IMO we should delete all the old stuff from the `ci-cd` folder and just keep these new pages.

fulyauluturk:
I agree, let me first check these and try to install CI/CD one more time and we'll have a final decision.

mkdocs.yml
```diff
@@ -136,15 +136,11 @@ nav:
...
-        - GitHub Actions: synthetics/ci-cd/github-actions.md
-        - Bitbucket Pipelines: synthetics/ci-cd/bitbucket-pipelines.md
-        - GitLab CI/CD: synthetics/ci-cd/gitlab-ci-cd.md
-        - Vercel: synthetics/ci-cd/vercel.md
-        - Netlify: synthetics/ci-cd/netlify.md
+        - Overview: synthetics/ci-cd/ci-cd-overview.md
+        - Installation: synthetics/ci-cd/ci-cd-installation.md
+        - Simple Workflow Example: synthetics/ci-cd/ci-cd-example-simple.md
+        - Complex Workflow Example: synthetics/ci-cd/ci-cd-example-complex.md
+        - Optional Check Run Fix: synthetics/ci-cd/ci-cd-check-run-fix.md
```
fulyauluturk, line 143: @HSSalman The pages we've removed are also listed under Integrations, so we need to remove them from there as well.

fulyauluturk (approved):

mkdocs.yml
HSSalman: Nice catch, I missed that completely. Fixed.

mkdocs.yml
```diff
@@ -210,15 +206,7 @@ nav:
...
-          - Overview: synthetics/ci-cd/overview.md
-          - Jenkins: synthetics/ci-cd/jenkins.md
-          - Travis CI: synthetics/ci-cd/travis-ci.md
-          - Circle CI: synthetics/ci-cd/circle-ci.md
-          - GitHub Actions: synthetics/ci-cd/github-actions.md
-          - Bitbucket Pipelines: synthetics/ci-cd/bitbucket-pipelines.md
-          - GitLab CI/CD: synthetics/ci-cd/gitlab-ci-cd.md
-          - Vercel: synthetics/ci-cd/vercel.md
-          - Netlify: synthetics/ci-cd/netlify.md
+          - GitHub Actions: synthetics/ci-cd/ci-cd-overview.md
```
fulyauluturk, line 209: @HSSalman Can we change ci-cd-overview.md to overview.md? This way, we won’t have to update the links in other places where we might have referred to it, such as in old blog posts or in SC.

HSSalman: Done

PR #684, opened 2025-06-20
ful_SC-22010_CICD_FAQ_ENTRIES

fulyauluturk:


otisg:
Because these are only about CICDMs, and CICDMs are a relatively small/narrow part of Synthetics, I would have maybe put these FAQs somewhere close/under CICDMs, but if this feels better to you then OK. CC @HSSalman

fulyauluturk:
Hmm you mean a dedicated FAQ page for these under CI/CD Integration section. Sure, why not.

PR #726, opened 2026-05-27
Delete docs/logs/mappings-templates.md

otisg:
There is no need to support direct index mapping manipulation, so we are removing this page.


otisg:
I hope I found all links to the removed page. I'll do another search/scan after this is merged and published.

fulyauluturk (approved):

PR #728, opened 2026-06-09
Fix search API hostname: logsene-receiver -> logsene-search

madmuffin1:
All examples on the **Using Sematext API** (search) page point at the `logsene-receiver.*` host. That host is ingestion-only and rejects read/search operations with HTTP 410:

```json
{"error":"Not supported","cause":"Logs receiver is a log ingestion endpoint only. Use the Logs Search Receiver ...","status":410}
```

Search/query/aggregate — and the real-time GET / `_mget` / `_msearch` read operations documented on this page — must target the `logsene-search.*` host instead.

Verified against a live US Logs App:

```
GET logsene-search.sematext.com/<token>/_search   -> 200
GET logsene-receiver.sematext.com/<token>/_search -> 410
```

This PR updates all 8 occurrences (US + EU hostnames) in `docs/logs/search-through-the-sematext-api.md` from `logsene-receiver` to `logsene-search`. No other content changed.

### DIFF
--- a/docs/logs/search-through-the-sematext-api.md
+++ b/docs/logs/search-through-the-sematext-api.md
@@ -6,7 +6,7 @@ You can also use this API to [index events or change the mapping](/docs/logs/ind
 
 When you use the API, here are the things you need to know:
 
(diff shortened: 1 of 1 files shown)


otisg:
Thanks @madmuffin1 this is all correct, we'll get this all updated pronto.

PR #732, opened 2026-07-10
SC-24603 Initial docs for the Sematext Cloud MCP Server

HSSalman:
This PR adds the initial docs for the Sematext Cloud MCP Server. I purposefully wanted to keep this short and sweet compared to our internal docs to not scare people away with a wall of text, and I believe the current form contains all the important information that a user could need. Sadly not a whole lot to screenshot so it'll be text-only, unless we want to create some graphics on how this works.

The `Tools` section of the overview feels a bit overkill, but it's at the bottom of the page so it shouldn't matter much. If you agree on it being unnecessary then we can get rid of it.

We'll also need to add the specific setup guides for the various AI agent providers later on down the line. I'll mention this during the demos and ping the volunteers to get the info and add it to the PR before we ship it.

### DIFF
--- a/docs/mcp/getting-started.md
+++ b/docs/mcp/getting-started.md
@@ -1,0 +1,54 @@
+title: Getting Started
+description: Installation instructions for the Sematext Cloud MCP Server
+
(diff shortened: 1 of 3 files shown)


otisg (approved):
Minimalistic. OK. LGTM.

PR #741, opened 2026-08-03
SC-25832: add go docs

abadojack:


HSSalman (commented):
Since this is Markdown, all text within a single paragraph should be on the same line, like elsewhere in the docs. Aside from that I don't have much to say since I haven't had much contact with this.

